PR #7, opened 2016-10-26
Blacklist the SaveV2 and RestoreV2 ops which need Tensor lists.

judah:


judah:
It looks like these ops were added in GitHub only recently:
https://github.com/tensorflow/tensorflow/commit/f9a80692963869ce649581ff1743d00fca5cd0c6

PR #8, opened 2016-10-26
Add instructions for building on Mac OS X.

judah:


README.md
```diff
@@ -22,3 +22,36 @@ There is also a demo application:
...
+
+- Build the TensorFlow library and install it on your machine:
+
+        cd third_party/tensorflow
+        ./configure  # Choose the defaults when prompted
+        bazel build -c opt tensorflow:libtensorflow_c.so
+        cp bazel-bin/tensorflow/libtensorflow_c.so /usr/local/lib
+        cd ../..
+
+- Append the following lines to this project's `stack.yaml`:
```
blackgnezdo: We should just do this. Our docker image relies on it too.

README.md
```diff
@@ -22,3 +22,36 @@ There is also a demo application:
...
+
+- Install dependencies via [Homebrew](http://brew.sh):
+
+        brew install swig
+        brew install bazel
+
+- Build the TensorFlow library and install it on your machine:
+
+        cd third_party/tensorflow
+        ./configure  # Choose the defaults when prompted
```
blackgnezdo: third_party/tensorflow/tensorflow/tools/docker/Dockerfile.devel
does this
tensorflow/tools/ci_build/builds/configured CPU

I wonder which one is right?

judah: That script just calls ./configure, and its comments suggest it's a hack specific to Docker.  So I think `./configure` is more appropriate here.  (It's also what the other language bindings recommend.)

README.md
judah: Done.

PR #11, opened 2016-10-27
Starting NN library

### LINKED ISSUE #10: missing a bunch of the "classification" functions
from here: https://www.tensorflow.org/versions/r0.11/api_docs/python/nn.html#classification

i.e. i'd like `sigmoid_cross_entropy_with_logits`, it's not there

glancing around it appears that there are many ops that are in `tensorflow/python/ops/nn.py` that aren't defined here; i guess because they can't be straightfowardly gen'd like the ones in `nn_ops.py`.

do the maintainers have some sort of plan for these? or should I just go about adding them as PRs as i happen on ones I care about?

silky:
- Added "sigmoidCrossEntropyWithLogits"
- Ported across a single test

fixes #10 (well, it adds a single function, and it's a place to start.)

the test could probably be cleaned up quite a bit; happy to do that later on, just wanted to get this up.

### DIFF
--- a/tensorflow-nn/src/TensorFlow/NN.hs
+++ b/tensorflow-nn/src/TensorFlow/NN.hs
@@ -19,17 +19,61 @@ module TensorFlow.NN
     ( sigmoidCrossEntropyWithLogits
     ) where
 
-import Prelude hiding (log, exp)
-import TensorFlow.Build
-import TensorFlow.Tensor
-import TensorFlow.Types
-import TensorFlow.Ops
-import TensorFlow.GenOps.Core (greaterEqual, select, log, exp)
+import Prelude hiding           ( log
+                                , exp
+                                )
+import TensorFlow.Build         ( Build(..)
+                                , render
+                                , withNameScope
+                                )
+import TensorFlow.GenOps.Core   ( greaterEqual
+                                , select
+                                , log
+                                , exp
+                                )
+import TensorFlow.Tensor        ( Tensor(..)
+                                , Value(..)
+                                )
+import TensorFlow.Types         ( TensorType(..)
+                                , OneOf
+                                )
+import TensorFlow.Ops           ( zerosLike
+                                , add
+                                )
 
+-- | Computes sigmoid cross entropy given `logits`.
+--
+-- Measures the probability error in discrete classification tasks in which each
+-- class is independent and not mutually exclusive.  For instance, one could
+-- perform multilabel classification where a picture can contain both an elephant
+-- and a dog at the same time.
+--
+-- For brevity, let `x = logits`, `z = targets`.  The logistic loss is
+--
+--        z * -log(sigmoid(x)) + (1 - z) * -log(1 - sigmoid(x))
+--      = z * -log(1 / (1 + exp(-x))) + (1 - z) * -log(exp(-x) / (1 + exp(-x)))
+--      = z * log(1 + exp(-x)) + (1 - z) * (-log(exp(-x)) + log(1 + exp(-x)))
+--      = z * log(1 + exp(-x)) + (1 - z) * (x + log(1 + exp(-x))
+--      = (1 - z) * x + log(1 + exp(-x))
+--      = x - x * z + log(1 + exp(-x))
+--
+--  For x < 0, to avoid overflow in exp(-x), we reformulate the above
+--
+--        x - x * z + log(1 + exp(-x))
+--      = log(exp(x)) - x * z + log(1 + exp(-x))
+--      = - x * z + log(1 + exp(x))
+--
+--  Hence, to ensure stability and avoid overflow, the implementation uses this
+--  equivalent formulation
+--
+--      max(x, 0) - x * z + log(1 + exp(-abs(x)))
+--
+--  `logits` and `targets` must have the same type and shape.
 sigmoidCrossEntropyWithLogits
-  :: (OneOf '[Float, Double] a, TensorType a, Num a) =>
-     Tensor Value a
-     -> Tensor Value a -> Build (Tensor Value a)
+  :: (OneOf '[Float, Double] a, TensorType a, Num a)
+     => Tensor Value a          -- ^ __logits__
+     -> Tensor Value a          -- ^ __targets__
+     -> Build (Tensor Value a)
 sigmoidCrossEntropyWithLogits logits targets = do
     let zeros = zerosLike logits
         cond = logits `greaterEqual` zeros
--- a/tensorflow-nn/src/TensorFlow/NN.hs
+++ b/tensorflow-nn/src/TensorFlow/NN.hs
@@ -73,13 +73,12 @@ sigmoidCrossEntropyWithLogits
   :: (OneOf '[Float, Double] a, TensorType a, Num a)
      => Tensor Value a          -- ^ __logits__
      -> Tensor Value a          -- ^ __targets__
-     -> Build (Tensor Value a)
+     -> Tensor Value a
 sigmoidCrossEntropyWithLogits logits targets = do
     let zeros = zerosLike logits
         cond = logits `greaterEqual` zeros
         relu_logits = select cond logits zeros
         neg_abs_logits = select cond (-logits) logits
-    withNameScope "logistic_loss" $ do
-        left  <- render $ relu_logits - logits * targets
-        right <- render $ log (1 + exp neg_abs_logits)
-        withNameScope "sigmoid_add" $ render $ left `add` right
+        left = relu_logits - logits * targets
+        right = log (1 + exp neg_abs_logits)
+    left `add` right
--- a/tensorflow-nn/src/TensorFlow/NN.hs
+++ b/tensorflow-nn/src/TensorFlow/NN.hs
@@ -22,10 +22,6 @@ module TensorFlow.NN
 import Prelude hiding           ( log
                                 , exp
                                 )
-import TensorFlow.Build         ( Build(..)
-                                , render
-                                , withNameScope
-                                )
 import TensorFlow.GenOps.Core   ( greaterEqual
                                 , select
                                 , log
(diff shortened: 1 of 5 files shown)


tensorflow-nn/src/TensorFlow/NN.hs
```diff
@@ -0,0 +1,41 @@
...
+
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module TensorFlow.NN
+    ( sigmoidCrossEntropyWithLogits
+    ) where
+
+import Prelude hiding (log, exp)
+import TensorFlow.Build
```
blackgnezdo: We tend to either import qualified (as TF) or enumerate the imports. This really improves readability. Could you do that?

tensorflow-nn/src/TensorFlow/NN.hs
```diff
@@ -0,0 +1,41 @@
...
+import TensorFlow.Tensor
+import TensorFlow.Types
+import TensorFlow.Ops
+import TensorFlow.GenOps.Core (greaterEqual, select, log, exp)
+
+sigmoidCrossEntropyWithLogits
```
blackgnezdo: Could you add a haddock? It should be OK to borrow from Python function doing the same since everything is copyright by TensorFlow authors.

tensorflow-nn/tests/NNTest.hs
```diff
@@ -0,0 +1,68 @@
...
+
+
+x, y :: [Float]
+x = [-100, -2, -2, 0, 2, 2,   2, 100]
+y = [   0,  0,  1, 0, 0, 1, 0.5,   1]
+shape = TF.Shape [8]
+
+
+logits, targets :: TF.Tensor TF.Value Float
+logits  = TF.constant shape x
```
blackgnezdo: I think you could save a bit by using TF.vector (which will compute shape from your list length)

tensorflow-nn/tests/NNTest.hs
```diff
@@ -0,0 +1,68 @@
...
+losses :: (TF.TensorType a, Floating a, Ord a) => [a] -> [a] -> [a]
+losses x' y' = sigmoidXentWithLogits x' y'
+
+
+testLogisticOutput = testCase "testLogisticOutput" $ do
+    let loss    = TF.sigmoidCrossEntropyWithLogits logits targets
+        ourLoss = V.fromList (losses x y)
+    --
+    r <- TF.runSession . TF.buildAnd TF.run $ loss
+    (all id (V.zipWith (\a b -> abs (a - b) <= 0.001) r ourLoss)) @? ("Xents too different: \n" ++ (show r) ++ "\n" ++ (show ourLoss))
```
blackgnezdo: Could you split this line? We try to stay below 80 chars.

tensorflow-nn/tests/NNTest.hs
```diff
@@ -0,0 +1,68 @@
...
+import Test.Framework.Providers.HUnit   (testCase)
+import Test.HUnit                       ((@?))
+
+import qualified Data.Vector            as V
+import qualified TensorFlow.Build       as TF
+import qualified TensorFlow.Ops         as TF
+import qualified TensorFlow.Session     as TF
+import qualified TensorFlow.Tensor      as TF
+import qualified TensorFlow.Types       as TF
+import qualified TensorFlow.NN          as TF
```
blackgnezdo: Maybe sort alphabetically?

tensorflow-nn/tests/NNTest.hs
```diff
@@ -0,0 +1,68 @@
...
+
+
+sigmoidXentWithLogits :: Floating a => Ord a => [a] -> [a] -> [a]
+sigmoidXentWithLogits logits' targets' =
+    let pred  = map (\x -> 1 / (1 + exp (-x))) logits'
+        eps   = 0.0001
+        pred' = map (\p -> min (max p eps) (1 - eps)) pred
+        f y z = (-z) * (log y) - (1 - z) * log (1 - y)
```
blackgnezdo: This looks like cross entropy, if so, give f a more informative name?

tensorflow-nn/tests/NNTest.hs
```diff
@@ -0,0 +1,68 @@
...
+     in zipWith f pred' targets'
+
+
+x, y :: [Float]
+x = [-100, -2, -2, 0, 2, 2,   2, 100]
```
blackgnezdo: Are x and y used just once? If so inline them?

tensorflow-nn/tests/NNTest.hs
```diff
@@ -0,0 +1,68 @@
...
+y = [   0,  0,  1, 0, 0, 1, 0.5,   1]
+shape = TF.Shape [8]
+
+
+logits, targets :: TF.Tensor TF.Value Float
+logits  = TF.constant shape x
+targets = TF.constant shape y
+
+
+losses :: (TF.TensorType a, Floating a, Ord a) => [a] -> [a] -> [a]
```
blackgnezdo: Looks like used just once, inline there (especially since sigmoidXentWithLogits is a much more informative name than losses)

blackgnezdo:
Overall: great! Thanks a lot for the PR and especially for the test.

You followed our style really closely which is pretty hard considering we didn't publish it :)

A bunch of nits in the test are just that: nits. Just trying to keep the code in the best possible shape.

silky:
i think i've address all the comments, and i've added some more tests.

however, as-is, one test, the `testGradientAtZero`, fails. the resulting gradients are `[0,0]`. i'm not sure why at the moment, but it will hopefully be apparent to me in the morning!

so don't merge till that's fixed!

silky:
okay; that failing test is fixed.

i just made `sigmoidCrossEntropyWithLogits` not return a `Build ...`. 

if there was a nicer way to fix that, then let me know! otherwise, i believe this is ready!

tensorflow-nn/src/TensorFlow/NN.hs
```diff
@@ -73,13 +73,12 @@ sigmoidCrossEntropyWithLogits
...
      => Tensor Value a          -- ^ __logits__
      -> Tensor Value a          -- ^ __targets__
-     -> Build (Tensor Value a)
+     -> Tensor Value a
 sigmoidCrossEntropyWithLogits logits targets = do
     let zeros = zerosLike logits
         cond = logits `greaterEqual` zeros
         relu_logits = select cond logits zeros
         neg_abs_logits = select cond (-logits) logits
-    withNameScope "logistic_loss" $ do
```
blackgnezdo: @fkm3, could you take a look at why Build is not behaving here? I know we eventually should have a non-Build story for assigning scopes, but until then, it would be nice to not have to choose between good names and correct code :)

tensorflow-nn/tests/NNTest.hs
```diff
@@ -16,53 +16,101 @@
...
+testLogisticOutputMultipleDim =
+        testCase "testLogisticOutputMultipleDim" $ do
+    let inputs   = defInputs 
+        shape    = [2, 2, 2]
+        vLogits  = TF.constant shape (logits  inputs)
+        vTargets = TF.constant shape (targets inputs)
+        tfLoss   = TF.sigmoidCrossEntropyWithLogits vLogits vTargets
+        ourLoss  = V.fromList $ sigmoidXentWithLogits (logits inputs) (targets inputs)
+    --
+    r <- TF.runSession . TF.buildAnd TF.run $ tfLoss
```
blackgnezdo: Looks like we implicitly flatten the result from [2,2,2] to [8] tensor here. Is this our long term plan @judah @fkm3 ?

judah: We should allow using a better type than Vector for this Fetch.  I haven't thought about it in detail, though.  Filed #13.

tensorflow-nn/src/TensorFlow/NN.hs
```diff
@@ -0,0 +1,84 @@
...
+{-# LANGUAGE OverloadedStrings #-}
+
+module TensorFlow.NN
+    ( sigmoidCrossEntropyWithLogits
+    ) where
+
+import Prelude hiding           ( log
+                                , exp
+                                )
+import TensorFlow.Build         ( Build(..)
```
silky: done

tensorflow-nn/src/TensorFlow/NN.hs
fkm3: Here is a version that works with `Build`:

```haskell
sigmoidCrossEntropyWithLogits
  :: (OneOf '[Float, Double] a, TensorType a, Num a)
     => Tensor Value a          -- ^ __logits__
     -> Tensor Value a          -- ^ __targets__
     -> Build (Tensor Value a)
sigmoidCrossEntropyWithLogits logits targets = do
    logits' <- render logits
    targets' <- render targets
    let zeros = zerosLike logits'
        cond = logits' `greaterEqual` zeros
        relu_logits = select cond logits' zeros
        neg_abs_logits = select cond (-logits') logits'
    withNameScope "logistic_loss" $ do
        left  <- render $ relu_logits - logits' * targets'
        right <- render $ log (1 + exp neg_abs_logits)
        withNameScope "sigmoid_add" $ render $ left `add` right
```

(the only difference is that it renders the input tensors outside of the scope)

This is one of the confusing aspects of how we are handling name scopes right now. The `logits` here is rendered inside the "logistic_loss" name scope and the one in the test (`vLogits` in `TF.gradients tfLoss [vLogits]`) is rendered outside of it, so you end up with two distinct tensors. The one in the test is never used, so the gradient w.r.t. it is always zero.

@blackgnezdo probably didn't run into this because we usually only calculate for the gradient w.r.t. a variable and variable creation is done inside `Build`.

blackgnezdo:
@silky, I leave up to you to make a call about which way you want to go: Build or no Build? I can be convinced either way...

silky:
cheers @blackgnezdo  and @fkm3 

i reverted it to the `Build ...` style, because that let's the implementation agree more with the one in the python code, which seems like a pretty good thing to aim for.

blackgnezdo:
Agreed, we should have better story around that Build consistency.

Thanks again for pushing through with this!

PR #14, opened 2016-10-28
Integration test

blackgnezdo:


ci_build/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Image for launching TensorFlow Haskell tests.
+FROM tensorflow/tensorflow:nightly-devel
+# gcr.io/tensorflow/tensorflow:latest-devel
+MAINTAINER TensorFlow authors <tensorflow-haskell@googlegroups.com>
+
+RUN \
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 575159689BEFB442 && \
+    echo 'deb http://download.fpcomplete.com/ubuntu trusty main'| tee /etc/apt/sources.list.d/fpco.list
+
+RUN apt-get update
```
gunan: To optimize docker's caching, I think it would be best to merge these commands under a single "RUN" command.

ci_build/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+        libcurl3-dev \
+        stack
+
+# The build context directory is the top of the tensorflow-haskell
+# tree.
+ADD . /tfhs
+
+WORKDIR /tfhs
+
+RUN git submodule init && git submodule update
```
gunan: Same here, any reason to not merge this to the above command?

ci_build/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Installs protoc and the libraries.
+RUN \
+    cd third_party/tensorflow && \
```
gunan: the nightly devel docker image should already have tensorflow installed on it under /workspace
do we need to fetch it separately?

blackgnezdo: The goal was to make the build more hermetic as tensorflow submodule checkout is pinned.

ci_build/Dockerfile
blackgnezdo: Done

ci_build/Dockerfile
blackgnezdo: Done

blackgnezdo:
I tried to reply the "why another Dockerfile" question in the comment. See if it suffices.

PR #17, opened 2016-11-07
Add TensorFlow.Core module to start formalizing the exposed API

fkm3:


fkm3:
Maybe this shouldn't re-export the ops modules and maybe TensorFlow.Ops should re-export CoreOps? Let me know what you think.

fkm3:
PTAL. It no longer re-rexports the op modules and has been moved to the `tensorflow` package.

I took a shot at improving the haddocks so that this module can act as a one page reference.

tensorflow/src/TensorFlow/Core.hs
```diff
@@ -0,0 +1,96 @@
...
+--
+-- Unless you are defining ops, you do not need to import other modules from
+-- this package.
+--
+-- Ops are provided in the following modules:
+--
+--     * "TensorFlow.Ops"
+--     * "TensorFlow.GenOps.Core"
+--     * "TensorFlow.Gradient"
+--     * "TensorFlow.EmbeddingOps"
```
fkm3: I'm not sure how to reference module in other packages. This doesn't work.

blackgnezdo: I wonder if this suggests we should change layering somewhat:

1. rename "tensorflow" into "tensorflow-base" (or -core)
2. rename "tensorflow-ops" to "tensorflow" and put this module there.
3. rename this module to TensorFlow

fkm3: That is what I was thinking originally, but if this module only re-exports stuff from "tensorflow-base", then I think we should keep it in the same package. I can change to this say something like "Ops are provided in the tensorflow-ops package" instead if there is no way to fix the haddock (also it may just be that I'm building the docs incorrectly).

fkm3: Updated it to refer to the cabal packages instead of the modules. Added the docs in another branch: https://github.com/tensorflow/haskell/tree/simple_imports_docs

(I rebased onto HEAD without thinking, so the history of the PR is a little messed up)

PR #18, opened 2016-11-08
Support ResourceHandle.

blackgnezdo:
Exposed by moving to newer TF.

### DIFF
--- a/tensorflow-opgen/src/TensorFlow/OpGen.hs
+++ b/tensorflow-opgen/src/TensorFlow/OpGen.hs
@@ -25,10 +25,11 @@ module TensorFlow.OpGen
 
 import Prelude hiding (head, tail)
 
+import Control.Applicative ((<**>))
 import Control.Monad (guard)
 import Data.Char (toLower, toUpper)
 import Data.Foldable (toList)
-import Data.Maybe (fromMaybe, maybeToList)
+import Data.Maybe (catMaybes, fromMaybe, maybeToList)
 import Data.ProtoLens (def, showMessage)
 import Data.List.NonEmpty (NonEmpty((:|)), head)
 import qualified Data.List.NonEmpty as NE
@@ -119,6 +120,10 @@ docOpList flags opList =
         , "{-# LANGUAGE OverloadedStrings #-}"
         , "{-# LANGUAGE RankNTypes #-}"
         , "{-# LANGUAGE ScopedTypeVariables #-}"
+          -- Avoids reports about shadowing standard library names.
+        , "{-# OPTIONS_GHC -fno-warn-name-shadowing #-}"
+          -- eqLengthGuard never returns false and dies instead.
+        , "{-# OPTIONS_GHC -fno-warn-incomplete-patterns #-}"
         , "module" <+> strictText moduleName <+> "where"
         , empty
         , imports
@@ -135,6 +140,7 @@ docOpList flags opList =
         shortName = Text.pack (takeBaseName $ outputFile flags)
         exclusions = Text.splitOn "," $ Text.pack $ excludeList flags
 
+camelCase :: Text -> Text
 camelCase s = Text.concat $ map upCase
                           $ filter (/= "ops")
                           $ Text.splitOn "_" s
@@ -152,6 +158,7 @@ forceCase :: (Char -> Char) -> Text -> Text
 forceCase convert s = maybe "" (\(c, cs) -> Text.cons (convert c) cs)
                       (Text.uncons s)
 
+imports :: Doc
 imports = stack [
       "import Data.ByteString (ByteString)"
     , "import Data.Complex (Complex)"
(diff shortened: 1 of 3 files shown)


blackgnezdo:
I'm yet to write a test, soliciting input early to check my direction.

tensorflow/src/TensorFlow/Output.hs
```diff
@@ -154,3 +155,8 @@ instance IsString Output where
         _ -> Output 0 $ assigned s
         where assigned n = Rendered $ def & name .~ Text.pack n
 
+
+-- | Represents an opaque handle to a mutable resource in the graph.
+-- Typical such resources are variables. The type parameter
+-- corresponds to the dtype of the tensor held in the variable.
+data ResourceHandle dtype = ResourceHandle Output
```
judah: newtype?  Also maybe 'a' instead of 'dtype' in the signature to match the other datatypes in this file; up to you.

tensorflow/src/TensorFlow/BuildOp.hs
```diff
@@ -77,11 +80,17 @@ instance ( OpResult a1
                <*> toResult
 
 tensorResult :: TensorKind v -> Result (Tensor v a)
-tensorResult v = do
+tensorResult v = (Tensor v) <$> recordResult
```
judah: Remove parens around "Tensor v"

tensorflow-opgen/src/TensorFlow/OpGen.hs
```diff
@@ -237,40 +245,57 @@ quotedText = dquotes . strictText
...
-        if Text.null (arg ^. numberAttr) then typ else brackets typ
+        if Text.null (arg ^. numberAttr) then typ else brackets <$> typ
       where typ = tensorType refType arg
-    tensorType refType arg =
-      "Tensor" <+> refType <+> maybe directType strictText indirectType
+    -- The result is (reference type variable if any, type representing the arg)
+    tensorType :: Doc -> OpDef'ArgDef  -> (Maybe Doc, Doc)
+    tensorType refType arg
+        -- Deals with resource handles that are unlike tensors and
+        -- have their own type level representation. The magic "dtype"
```
judah: Can you point me to the docs (or code) referencing the magic "dtype" name?  I couldn't find it.

tensorflow/src/TensorFlow/BuildOp.hs
```diff
@@ -13,6 +13,9 @@
 -- limitations under the License.
 
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE KindSignatures #-}
```
judah: I'm suspicious whether any of these added extensions are actually needed for this change in this file :)

tensorflow-opgen/src/TensorFlow/OpGen.hs
```diff
@@ -237,40 +245,57 @@ quotedText = dquotes . strictText
...
                          (\(t, aDef) ->
                            "TensorType" <+> strictText t
                            : maybeToList (oneOfRestrictions aDef t))
                          (Map.elems typeMap)) <+> "=>"
     tensorInputs = zipWith tensorArg refTypes (d ^. inputArg)
-    refTypes = map (\x -> "v" <> int x) [1..length (d ^. inputArg)]
-    tensorArg refType arg = wrapArg refType arg <+>
-                            hang 0 ("-- ^" <+> argComment arg)
+    refTypes = ["v" <> int x | x <- [1..length (d ^. inputArg)]]
+    refVariableNames = catMaybes (map fst tensorInputs)
```
fkm3: Maybe move `refVariableNames` above `tensorInputs`? They were in a kind of reversed "dependency" order before and now it is a bit jumbled.

tensorflow-opgen/src/TensorFlow/OpGen.hs
fkm3: It seems like this convention will break if there is an op that accepts two resource handles with different types.

tensorflow/src/TensorFlow/BuildOp.hs
blackgnezdo: Not anymore. They were when recordResult was different.

tensorflow/src/TensorFlow/Output.hs
blackgnezdo: Done

tensorflow/src/TensorFlow/Output.hs
```diff
@@ -154,3 +155,8 @@ instance IsString Output where
         _ -> Output 0 $ assigned s
         where assigned n = Rendered $ def & name .~ Text.pack n
 
+
+-- | Represents an opaque handle to a mutable resource in the graph.
+-- Typical such resources are variables. The type parameter
+-- corresponds to the dtype of the tensor held in the variable.
```
blackgnezdo: Done

tensorflow-opgen/src/TensorFlow/OpGen.hs
blackgnezdo: Done

Issue #19, opened 2016-11-08
"embedding_lookup" appears to be broken, maybe missing some reshaping?


silky:
i've been trying to use this function in some model, but i think it's broken. i'm trying to fix it, but i might not have time today, so i want to note it.

there appear to be a few problems:

0. we could add the special-case for np = 1 that is in the original [python](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/embedding_ops.py#L91) (that would actually probably fix the problem i'm facing right now)

1. in [embeddingLookup](https://github.com/tensorflow/haskell/blob/67690d149986f3787d3768594ee544fca6ee1d83/tensorflow-ops/src/TensorFlow/EmbeddingOps.hs#L65) the `dynamicPartition` operation fails for me with:

```
TensorFlowException TF_INVALID_ARGUMENT: "data.shape must start with partitions.shape, got data.shape = [12000], partitions.shape = [100,120,1] ..."
```

i believe this is because we are missing reshaping ids to the ids to get the `flatIds` that is in the python.

2. we're missing a bunch of reshaping at the end

i believe this is just simply missing from the implementation in the haskell code here, compared to the python: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/embedding_ops.py#L160

maybe it's missing because we don't have any of the `framework/tensor_shape.py`  code.

i think to fix this we should probably add some tests. when i get some time i might port some of the python ones again, and hopefully identify some actual problems that way.

blackgnezdo:
Yes, quite possible that it's not entirely faithful. It was used in one project in a particular way. If you could create a breaking unit test it would help a lot :+1:

Issue #13, opened 2016-10-27
Fetch multi-dimensions arrays


judah:
Currently, the only options for fetching a multi-dimensional Tensor are lists or Vectors, both of which concatenate it into one dimension.  (For example, a shape of [2,2,2] will turn into a shape of [8]).  We should allow using a more appropriate Haskell type for such Fetches.

blackgnezdo:
Is it mostly predicated on us deciding what we want tensors to look like abstracted from TF?

PR #20, opened 2016-11-08
added failing test

silky:
see https://github.com/tensorflow/haskell/issues/19

this fails. i expect that it should work and that it should result in the given output and shape.

~~once this test works, i'll need to make sure it actually properly checks the shape (at this point this can only be done correctly as a tensorflow op, because we can't yet get back multi-dim vectors https://github.com/tensorflow/haskell/issues/13)~~

### DIFF
--- a/tensorflow-ops/src/TensorFlow/EmbeddingOps.hs
+++ b/tensorflow-ops/src/TensorFlow/EmbeddingOps.hs
@@ -60,9 +60,12 @@ embeddingLookup :: forall a b v .
                 -- fewer than 2^31 entries.
                 -> Build (Tensor Value a)
                 -- ^ A dense tensor with shape `shape(ids) + shape(params)[1:]`.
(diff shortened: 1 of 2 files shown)


blackgnezdo:
Jenkins test this please.

silky:
as a side note, this other test that fails, the `Op TopK is not available in GraphDef version 15. It has been removed in version 7. Use TopKV2 instead` ... could probably be fixed. i'm not sure if you all see that when you run the tests locally, but i see it only when i have another test failing; it mysteriously goes away when it's the only test that fails.

blackgnezdo:
Thanks for the test case! :+1:

I also just now noticed the TopKV2 issue. Let's see if we can crush these.

silky:
i added a new commit to the source fixing the case when there is only a single partition

i also added a test that demonstrates that the multi-partition case is still broken

PR #22, opened 2016-11-09
Embedding lookup fix

blackgnezdo:
@silky I added a fix on top of your commit. Care to review?

### DIFF
--- a/tensorflow-ops/src/TensorFlow/EmbeddingOps.hs
+++ b/tensorflow-ops/src/TensorFlow/EmbeddingOps.hs
@@ -23,9 +23,8 @@ module TensorFlow.EmbeddingOps where
 
 import Control.Monad (zipWithM)
 import Data.Int (Int32, Int64)
-import Data.List (genericLength)
 import TensorFlow.Build (Build, colocateWith, render)
-import TensorFlow.Ops (shape, scalar, vector)  -- Also Num instance for Tensor
+import TensorFlow.Ops (scalar, shape, vector)  -- Also Num instance for Tensor
 import TensorFlow.Tensor (Tensor, Value)
 import TensorFlow.Types (OneOf, TensorType)
 import qualified TensorFlow.GenOps.Core as CoreOps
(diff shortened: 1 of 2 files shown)


tensorflow-ops/src/TensorFlow/EmbeddingOps.hs
```diff
@@ -56,21 +56,33 @@ embeddingLookup :: forall a b v .
...
+                        params gatherIds
+    let unshapedResult = CoreOps.dynamicStitch pindices partitionedResult
+    -- Shape restoration is not as optimal as it would be with client
+    -- side shape tracking.
+    paramShape <- colocateWith p0 (render (shape p0))
+    let finalShape = CoreOps.concat 0 [shape ids, tailShape]
+        tailShape = CoreOps.slice paramShape (singleton 1) (singleton (-1))
+    render $ CoreOps.reshape unshapedResult finalShape
+  where
+    np = genericLength params
```
fkm3:
```
genericLength           :: (Num i) => [a] -> i
genericLength []        =  0
genericLength (_:l)     =  1 + genericLength l
```

I think `0`, `1`, and `(+)` are all going to use the Tensor instance, which means this fold is evaluated in the tensorflow graph. A bit scary...

blackgnezdo: Nice catch! Fixed.

blackgnezdo:
@googlebot the code was authored by @silky who has previously contributed and signed the CLA. He approved this PR.

tensorflow-ops/src/TensorFlow/EmbeddingOps.hs
silky: Are you sure? There's no constraint on `a`, which is the Tensor, so the `0` is bound by the `Num` constraint on `i`, no?

(furthermore, even a `Num a` requirement on genericLength doesn't work, because it can't guarantee that the `v`  on params :: Tensor v a is  `Value` ...)

fkm3: `np` is used as an input to `CoreOps.mod`, so it must be a tensor, which makes the `i` in `genericLength` a tensor.

PR #23, opened 2016-11-10
test case to show can't calculate grad for embedding (and associated fix)

silky:
i've noticed that i don't seem to be able to calculate the gradient w.r.t an embedding, this is the error:

```
ERROR: TensorFlowException TF_INVALID_ARGUMENT "Expected begin and size arguments to be 1-D tensors of size 1, but got shapes [] and [] instead.

 [[Node: Slice_54 = Slice[Index=DT_INT32, T=DT_INT32, _device=\"/job:localhost/replica:0/task:0/cpu:0\"](Shape_52, Const_18, Neg_53)]]"
```

the equivalent python code to the following test works (but i note that in python i can *also* calculate the gradient w.r.t say the variable named below called `op` that is a `Tensor Value` not a `Tensor Ref`; i.e. it's a type error)

i'm looking for a bit of guidance here. i'm fairly sure i should be able to calculate the gradient like this. but i could be convinced/shown that i'm doing it wrong, and there's a different way to get the embedding vars to gradient application ...

interested in thoughts.

if we do keep this test, i should first make a bit of a test lib and put some funtions in there (namely the `assertAllClose`...)

(sidenote: i'm having a truly terrible time with shape instances/Int32s/etc, so i'll one day try and figure out how to clean all that up; maybe i'm just missing some imports or some philosophical viewpoint)

### DIFF
--- a/tensorflow-ops/src/TensorFlow/Gradient.hs
+++ b/tensorflow-ops/src/TensorFlow/Gradient.hs
@@ -407,13 +407,13 @@ toT = Tensor ValueKind
 -- | Wrapper around `TensorFlow.GenOps.Core.slice` that builds vectors from scalars for
 -- simple slicing operations.
 flatSlice :: forall v1 t i . (TensorType t)
-         => Tensor v1 t -- ^ __input__
-         -> Int32 -- ^ __begin__: specifies the offset into the first dimension of
-                            -- 'input' to slice from.
-         -> Int32 -- ^ __size__: specifies the number of elements of the first dimension
-                            -- of 'input' to slice. If size is -1, all remaining elements in the dimension
-                            -- are included in the slice (i.e. this is equivalent to setting
-                            -- size = input.dim_size(0) - begin).
+         => Tensor v1 t    -- ^ __input__
+         -> Int32          -- ^ __begin__: specifies the offset into the first dimension of
+                           -- 'input' to slice from.
+         -> Int32          -- ^ __size__: specifies the number of elements of the first dimension
+                           -- of 'input' to slice. If size is -1, all remaining elements in the dimension
+                           -- are included in the slice (i.e. this is equivalent to setting
+                           -- size = input.dim_size(0) - begin).
          -> Tensor Value t -- ^ __output__
 flatSlice input begin size = CoreOps.slice input (vector [begin]) (vector [size])
 
@@ -446,9 +446,9 @@ opGrad "Gather" _ [toT -> x, toT -> indices] [dz] =
   where
     -- TODO(gnezdo): Use colocateWith but it requires Build monad.
     denseShape = shape (x :: Tensor Value a)
-    numRows = scalarize $ flatSlice denseShape (0 :: Int32) 1
+    numRows = scalarize $ flatSlice denseShape 0 1
     valuesShape = CoreOps.concat 0 [ allDimensions
-                                   , flatSlice denseShape 1 (-1 :: Int32)
+                                   , flatSlice denseShape 1 (-1)
                                    ]
     values = reshape dz valuesShape
     -- TODO(fmayle): This could be either Int32 or Int64.
@@ -643,7 +643,7 @@ opGrad "SparseSegmentSum" _ [toT -> x, toT -> y, toT -> t] [dz] =
     , Nothing
     , Nothing
     ]
-  where inputRows = flatSlice (shape (x :: Tensor Value a)) (0 :: Int32) 1
+  where inputRows = flatSlice (shape (x :: Tensor Value a)) 0 1
 
 opGrad "LabelClasses" _ _ _ = [Nothing, Nothing]
 opGrad "LabelWeights" _ _ _ = [Nothing]
--- a/tensorflow-ops/src/TensorFlow/Gradient.hs
+++ b/tensorflow-ops/src/TensorFlow/Gradient.hs
@@ -404,6 +404,20 @@ toT :: Output -> Tensor Value a
 toT = Tensor ValueKind
 
 
+-- | Wrapper around `GenOps.Core.slice` that builds vectors from scalars for
+-- simple slicing operations.
+flatSlice :: forall v1 t i . (TensorType t, OneOf '[Int32, Int64] i)
+         => Tensor v1 t -- ^ __input__
+         -> i -- ^ __begin__: specifies the offset into the first dimension of
+                            -- 'input' to slice from.
+         -> i -- ^ __size__: specifies the number of elements of the first dimension
+                            -- of 'input' to slice. If size is -1, all remaining elements in the dimension
+                            -- are included in the slice (i.e. this is equivalent to setting
+                            -- size = input.dim_size(0) - begin).
+         -> Tensor Value t -- ^ __output__
+flatSlice input begin size = CoreOps.slice input (vector [begin]) (vector [size])
+
+
 -- | The gradient function for an op type.
 --
 -- These implementations should match their python counterparts in:
@@ -432,10 +446,11 @@ opGrad "Gather" _ [toT -> x, toT -> indices] [dz] =
   where
     -- TODO(gnezdo): Use colocateWith but it requires Build monad.
     denseShape = shape (x :: Tensor Value a)
-    numRows = toScalar $ CoreOps.slice denseShape (vector [0 :: Int32]) (vector [1])
+    numRows = toScalar $ flatSlice denseShape (0 :: Int32) 1
+    -- numRows = toScalar $ CoreOps.slice denseShape (vector [0 :: Int32]) (vector [1])
     valuesShape = CoreOps.concat 0 [
                                  allDimensions
-                               , CoreOps.slice denseShape (vector [1]) (vector [-1 :: Int32])
+                               , flatSlice denseShape 1 (-1 :: Int32)
                                ]
     values = reshape dz valuesShape
     -- TODO(fmayle): This could be either Int32 or Int64.
@@ -630,7 +645,7 @@ opGrad "SparseSegmentSum" _ [toT -> x, toT -> y, toT -> t] [dz] =
     , Nothing
     , Nothing
     ]
-  where inputRows = CoreOps.slice (shape (x :: Tensor Value a)) (vector [0 :: Int32]) (vector [1])
+  where inputRows = flatSlice (shape (x :: Tensor Value a)) (0 :: Int32) 1
 
 opGrad "LabelClasses" _ _ _ = [Nothing, Nothing]
 opGrad "LabelWeights" _ _ _ = [Nothing]
--- a/tensorflow-ops/src/TensorFlow/Gradient.hs
+++ b/tensorflow-ops/src/TensorFlow/Gradient.hs
@@ -77,6 +77,7 @@ import TensorFlow.Ops
     , shape
     , softmaxCrossEntropyWithLogits
     , sum
+    , toScalar
     , vector
     , zerosLike
     )
@@ -402,6 +403,7 @@ type GradientFunc a = NodeDef
 toT :: Output -> Tensor Value a
 toT = Tensor ValueKind
 
+
 -- | The gradient function for an op type.
 --
 -- These implementations should match their python counterparts in:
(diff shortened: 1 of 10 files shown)


blackgnezdo:
@fkm3 could you take a look if you have some time?

tensorflow-ops/tests/EmbeddingOpsTest.hs
```diff
@@ -85,6 +91,45 @@ testEmbeddingLookupHasRightShape = testCase "testEmbeddingLookupHasRightShape" $
     values @=? V.fromList [ 1, 1, 1, 0, 0, 0 ]
 
 
+-- | Check that we can calculate gradients w.r.t gradients.
```
blackgnezdo: The second gradients presumably should be "embeddings".

blackgnezdo:
@silky you are not missing anything anything deep around shapes and int32s. It's an artifact of TF having started migrating to int64. On Haskell side we should just move to int64 as soon there's enough infrastructure in C++. We have no legacy to worry about and types will help us find the code in need of a change.

@vrv, do you happen to know if int64 shape support is complete enough now to transition?

vrv:
No, you should be cc'ed on the internal bug that describes why this is hard. :/

Though I don't fully understand why it's relevant here: is it that the system can't handle being able to output tensors whose elements are either int32 or int64?

blackgnezdo:
@vrv, it can, we are just forced to choose Int32 or Int64. And I don't want to default to Int32 (which is the thing that works fully, right?) as the long term default will be Int64.

vrv:
Only thing I can suggest is that we coordinate the switch of the default from Int32 to Int64 if/when it happens.

fkm3:
The gradient code is using rank 0 tensors as parameters to `slice`, which is apparently illegal. I remember it working before...

I think we just need to update all of the calls to `CoreOps.slice` in https://github.com/tensorflow/haskell/blob/master/tensorflow-ops/src/TensorFlow/Gradient.hs

silky:
thanks @blackgnezdo, i've added that in.

blackgnezdo:
@silky, did you get to try @fkm3's suggestion yet?

silky:
@blackgnezdo  - nope, i might have some time to try it out this weekend but i'm not sure; i'll report back when i have a go

silky:
@fkm3 @blackgnezdo  - yeah, that basically fixed it; i also needed to convert the last argument to a scalar. it works now!

please review, and i'm happy to create that test lib at some point in next day or so then merge.

silky:
i've added that lib now; let me know if anything else needs to be done!

tensorflow-ops/tests/EmbeddingOpsTest.hs
```diff
@@ -85,6 +92,36 @@ testEmbeddingLookupHasRightShape = testCase "testEmbeddingLookupHasRightShape" $
     values @=? V.fromList [ 1, 1, 1, 0, 0, 0 ]
 
 
+-- | Check that we can calculate gradients w.r.t embeddings.
+testEmbeddingLookupGradients = testCase "testEmbeddingLookupGradients" $ do
+    let xVals = V.fromList ([20, 20] :: [Float]) -- Agrees with "embedding", so gradient should be zero.
```
blackgnezdo: We try to keep lines under < 80 chars. Could you check if this line fits?

tensorflow-ops/tests/EmbeddingOpsTest.hs
```diff
@@ -85,6 +92,36 @@ testEmbeddingLookupHasRightShape = testCase "testEmbeddingLookupHasRightShape" $
...
+
+    gs <- TF.runSession $ do
+        grads <- TF.build $ do
+            let shape         = TF.Shape [2, 1] 
+            let embeddingInit = [1, 20] :: [Float]
+            let idValues      = [1, 1]   :: [Int32]
+            let ids           = TF.constant (TF.Shape [1, 2]) idValues
+
+            x <- TF.placeholder (TF.Shape [2]) 
+            embedding <- TF.initializedVariable =<< (TF.render $ TF.constant shape embeddingInit)
```
blackgnezdo: TF.render (TF.constant shape embeddingInit)  -- should work too?

tensorflow-ops/tests/EmbeddingOpsTest.hs
```diff
@@ -85,6 +92,36 @@ testEmbeddingLookupHasRightShape = testCase "testEmbeddingLookupHasRightShape" $
...
+-- | Check that we can calculate gradients w.r.t embeddings.
+testEmbeddingLookupGradients = testCase "testEmbeddingLookupGradients" $ do
+    let xVals = V.fromList ([20, 20] :: [Float]) -- Agrees with "embedding", so gradient should be zero.
+    let shape = TF.Shape [2]
+
+    gs <- TF.runSession $ do
+        grads <- TF.build $ do
+            let shape         = TF.Shape [2, 1] 
+            let embeddingInit = [1, 20] :: [Float]
+            let idValues      = [1, 1]   :: [Int32]
```
blackgnezdo: Remove extra spaces before ::?

tensorflow-ops/tests/EmbeddingOpsTest.hs
```diff
@@ -85,6 +92,36 @@ testEmbeddingLookupHasRightShape = testCase "testEmbeddingLookupHasRightShape" $
...
+
+            grad <- fmap head (TF.gradients loss [embedding])
+
+            return $ \xs -> TF.runWithFeeds [TF.feed x xs] grad
+
+        grads (TF.encodeTensorData shape xVals :: TF.TensorData Float)
+ 
+    -- Gradients should be zero (or close)
+    assertAllClose gs (V.fromList ([0, 0] :: [Float]))
+
```
blackgnezdo: Remove a couple of empty lines?

tensorflow-ops/tests/EmbeddingOpsTest.hs
```diff
@@ -85,6 +92,36 @@ testEmbeddingLookupHasRightShape = testCase "testEmbeddingLookupHasRightShape" $
...
+            let loss = TF.mean (CoreOps.square (TF.abs (op - x))) (TF.scalar (0 :: Int32))
+
+            grad <- fmap head (TF.gradients loss [embedding])
+
+            return $ \xs -> TF.runWithFeeds [TF.feed x xs] grad
+
+        grads (TF.encodeTensorData shape xVals :: TF.TensorData Float)
+ 
+    -- Gradients should be zero (or close)
+    assertAllClose gs (V.fromList ([0, 0] :: [Float]))
```
blackgnezdo: [0, 0 :: Float] should work too.

tensorflow-ops/src/TensorFlow/Ops.hs
```diff
@@ -256,6 +257,15 @@ constant (Shape shape') values
                       [def & TensorShape.size .~ x | x <- shape']
                 & tensorVal .~ values
 
+-- | Reshape a N-D tensor down to a scalar.
+toScalar :: (TensorType a)
```
blackgnezdo: This should probably be named scalarize or something similar (extractScalar?). We already have "unScalar".

tensorflow-ops/src/TensorFlow/Ops.hs
```diff
@@ -256,6 +257,15 @@ constant (Shape shape') values
                       [def & TensorShape.size .~ x | x <- shape']
                 & tensorVal .~ values
 
+-- | Reshape a N-D tensor down to a scalar.
+toScalar :: (TensorType a) 
+         => Tensor v a 
+         -> Tensor Value a
+toScalar = flip CoreOps.reshape (vector scalarShape)
```
blackgnezdo: I'd avoid flip here. 
toScalar x = CoreOps.reshape x (vector scalarShape)
Is both shorter and easier to understand.

tensorflow-ops/src/TensorFlow/Ops.hs
```diff
@@ -256,6 +257,15 @@ constant (Shape shape') values
                       [def & TensorShape.size .~ x | x <- shape']
                 & tensorVal .~ values
 
+-- | Reshape a N-D tensor down to a scalar.
```
blackgnezdo: Does it fail at runtime if the tensor has more than 1 element?

tensorflow-ops/src/TensorFlow/Ops.hs
```diff
@@ -256,6 +257,15 @@ constant (Shape shape') values
...
+toScalar :: (TensorType a) 
+         => Tensor v a 
+         -> Tensor Value a
+toScalar = flip CoreOps.reshape (vector scalarShape)
+    where
+        scalarShape = [] :: [Int32]
```
blackgnezdo: Maybe link to [reshape documentation](https://www.tensorflow.org/versions/r0.11/api_docs/python/array_ops.html#reshape)
which has an example?

tensorflow-ops/src/TensorFlow/Ops.hs
silky: ah; squeeze, i forgot about it, good suggestion!

silky: i've decided i'm torn

ideally one would have a `rank` associated to Tensors, and we could say that `scalarize` goes from `rank n -> rank 0`, whereas `squeeze` would go from `rank n -> rank m` with `m <= n`. and it might be a nice dream to one day have functions that demand `rank 0` tensors.

maybe that's a bit too demanding at this point though.

tensorflow-test/tensorflow-test.cabal
```diff
@@ -0,0 +1,25 @@
+name:                tensorflow-test
+version:             0.1.0.0
+synopsis:            Friendly layer around TensorFlow bindings.
```
blackgnezdo: Adjust to match intent?

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -402,6 +403,21 @@ type GradientFunc a = NodeDef
 toT :: Output -> Tensor Value a
 toT = Tensor ValueKind
 
+
+-- | Wrapper around `TensorFlow.GenOps.Core.slice` that builds vectors from scalars for
+-- simple slicing operations.
+flatSlice :: forall v1 t i . (TensorType t, OneOf '[Int32, Int64] i)
```
blackgnezdo: You seem to only need the Int32 version, let's reduce generality and improve call sites?

tensorflow-ops/tests/EmbeddingOpsTest.hs
```diff
@@ -23,32 +23,40 @@ import Data.List (genericLength)
...
 
+
 -- | Tries to perform a simple embedding lookup, with two partitions.
-testEmbeddingLookupHasRightShapeWithPartition = testCase "testEmbeddingLookupHasRightShapeWithPartition" $ do
+testEmbeddingLookupHasRightShapeWithPartition = 
+        testCase "testEmbeddingLookupHasRightShapeWithPartition" $ do
     let shape       = TF.Shape [1, 3] -- Consider a 3-dim embedding of two items.
-    let embedding1  = [ 1, 1, 1 ] :: [Int32]
-    let embedding2  = [ 0, 0, 0 ] :: [Int32]
+    let embedding1  = [ 1, 1, 1 :: Int32]
```
blackgnezdo: Do you feel like running this file through the latest version of hindent? I bet it's no fun to chase down all these space nits (extra one after [)

tensorflow-test/src/TensorFlow/Test.hs
```diff
@@ -0,0 +1,34 @@
...
+
+{-# LANGUAGE OverloadedStrings #-}
+
+module TensorFlow.Test
+    ( assertAllClose
+    ) where
+
+import Test.HUnit.Lang (Assertion(..))
+import qualified Data.Vector as V
+import Test.HUnit ((@?))
```
blackgnezdo: Sort imports.

blackgnezdo:
Jenkins please test this.

blackgnezdo:
ok to test

tensorflow-ops/tests/EmbeddingOpsTest.hs
silky: i tried all the available styles and they all missed with all the other indenting in fairly radical ways

i cleaned these up, and a few others. don't mind if you point out things i missed :)

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -402,6 +403,21 @@ type GradientFunc a = NodeDef
 toT :: Output -> Tensor Value a
 toT = Tensor ValueKind
 
+
+-- | Wrapper around `TensorFlow.GenOps.Core.slice` that builds vectors from scalars for
+-- simple slicing operations.
+flatSlice :: forall v1 t i . (TensorType t)
+         => Tensor v1 t -- ^ __input__
+         -> Int32 -- ^ __begin__: specifies the offset into the first dimension of
+                            -- 'input' to slice from.
```
blackgnezdo: Maybe align the -- with the ones above?

tensorflow-ops/src/TensorFlow/Ops.hs
```diff
@@ -256,6 +257,17 @@ constant (Shape shape') values
                       [def & TensorShape.size .~ x | x <- shape']
                 & tensorVal .~ values
 
+-- | Reshape a N-D tensor down to a scalar.
+-- 
+-- See `TensorFlow.GenOps.Core.reshape`.
+scalarize :: (TensorType a) 
+         => Tensor v a 
+         -> Tensor Value a
```
blackgnezdo: Probably the signature will fit on one line?

PR #27, opened 2016-11-13
Optimize fetching

fkm3:
# Before

```
benchmarking feedFetch/4 byte
time                 56.07 μs   (54.64 μs .. 57.31 μs)
                     0.995 R²   (0.993 R² .. 0.997 R²)
mean                 57.66 μs   (56.57 μs .. 58.74 μs)
std dev              3.851 μs   (3.350 μs .. 4.642 μs)
variance introduced by outliers: 68% (severely inflated)

benchmarking feedFetch/4 KiB
time                 228.8 μs   (227.4 μs .. 230.3 μs)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 229.1 μs   (228.5 μs .. 230.5 μs)
std dev              3.016 μs   (1.902 μs .. 5.111 μs)

benchmarking feedFetch/4 MiB
time                 596.5 ms   (457.2 ms .. 661.2 ms)
                     0.994 R²   (0.984 R² .. 1.000 R²)
mean                 569.8 ms   (536.0 ms .. 589.7 ms)
std dev              30.62 ms   (0.0 s .. 34.45 ms)
variance introduced by outliers: 19% (moderately inflated)
```

# After

```
benchmarking feedFetch/4 byte
time                 58.26 μs   (57.03 μs .. 59.75 μs)
                     0.996 R²   (0.993 R² .. 0.998 R²)
mean                 61.08 μs   (59.17 μs .. 65.55 μs)
std dev              9.397 μs   (4.998 μs .. 17.20 μs)
variance introduced by outliers: 93% (severely inflated)

benchmarking feedFetch/4 KiB
time                 96.54 μs   (96.08 μs .. 97.07 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 97.85 μs   (97.21 μs .. 98.96 μs)
std dev              2.595 μs   (1.807 μs .. 4.076 μs)
variance introduced by outliers: 23% (moderately inflated)

benchmarking feedFetch/4 MiB
time                 126.0 ms   (125.4 ms .. 127.3 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 127.5 ms   (127.1 ms .. 128.0 ms)
std dev              720.2 μs   (580.4 μs .. 882.0 μs)
variance introduced by outliers: 11% (moderately inflated)
```

### DIFF
--- a/tensorflow-ops/tensorflow-ops.cabal
+++ b/tensorflow-ops/tensorflow-ops.cabal
@@ -187,6 +187,14 @@ Test-Suite TypesTest
                , test-framework-quickcheck2
                , vector
 
+Benchmark FeedFetchBench
+  default-language: Haskell2010
+  type:       exitcode-stdio-1.0
+  main-is:    FeedFetchBench.hs
+  hs-source-dirs: tests
+  build-depends: base, criterion, tensorflow, tensorflow-ops, vector, transformers
+  ghc-options: -O2 -threaded
+
 source-repository head
   type:     git
   location: https://github.com/tensorflow/haskell
--- a/tensorflow-ops/tensorflow-ops.cabal
+++ b/tensorflow-ops/tensorflow-ops.cabal
@@ -192,7 +192,13 @@ Benchmark FeedFetchBench
   type:       exitcode-stdio-1.0
   main-is:    FeedFetchBench.hs
   hs-source-dirs: tests
-  build-depends: base, criterion, tensorflow, tensorflow-ops, vector, transformers
+  build-depends: base
+               , criterion
+               , deepseq
+               , tensorflow
+               , tensorflow-ops
+               , transformers
+               , vector
   ghc-options: -O2 -threaded
 
 source-repository head
--- a/tensorflow-ops/tests/FeedFetchBench.hs
+++ b/tensorflow-ops/tests/FeedFetchBench.hs
@@ -1,31 +1,25 @@
 -- Disable full-laziness to keep ghc from optimizing most of the benchmark away.
 {-# OPTIONS_GHC -fno-full-laziness #-}
+import Control.DeepSeq (NFData(rnf))
 import Control.Exception (evaluate)
-import Control.Monad (replicateM_)
 import Control.Monad.IO.Class (liftIO)
-import Criterion.Main (defaultMain, bgroup, bench, nfIO)
+import Criterion.Main (defaultMain, bgroup, bench)
 import Criterion.Types (Benchmarkable(..))
 import qualified Data.Vector as V
-import qualified TensorFlow.Build as TF
-import qualified TensorFlow.ControlFlow as TF
-import qualified TensorFlow.Gradient as TF
-import qualified TensorFlow.Nodes as TF
+import qualified TensorFlow.Core as TF
 import qualified TensorFlow.Ops as TF
-import qualified TensorFlow.Session as TF
-import qualified TensorFlow.Tensor as TF
-import qualified TensorFlow.Types as TF
 
 -- | Create 'Benchmarkable' for 'TF.Session'.
 --
 -- The entire benchmark will be run in a single tensorflow session. The
 -- 'TF.Session' argument will be run once and then its result will be run N
 -- times.
-whnfSession :: TF.Session (a -> TF.Session b) -> a -> Benchmarkable
-whnfSession init x = Benchmarkable $ \m -> TF.runSession $ do
+nfSession :: NFData b => TF.Session (a -> TF.Session b) -> a -> Benchmarkable
+nfSession init x = Benchmarkable $ \m -> TF.runSession $ do
     f <- init
     -- Can't use replicateM because n is Int64.
     let go n | n <= 0    = return ()
-             | otherwise = f x >>= liftIO . evaluate >> go (n-1)
+             | otherwise = f x >>= liftIO . evaluate . rnf >> go (n-1)
     go m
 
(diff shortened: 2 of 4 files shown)


fkm3:
FYI in case someone has peeked at this PR already: GHC was optimizing away most of the benchmark. I've fixed it and updated the numbers in the description.

fkm3:
I've got a more involved change that makes a bigger difference here: https://github.com/tensorflow/haskell/commit/f176806df63c71eeee486e7ee9f7c640257f89bf

tensorflow-ops/tests/FeedFetchBench.hs
```diff
@@ -0,0 +1,49 @@
+-- Disable full-laziness to keep ghc from optimizing most of the benchmark away.
+{-# OPTIONS_GHC -fno-full-laziness #-}
+import Control.Exception (evaluate)
+import Control.Monad (replicateM_)
+import Control.Monad.IO.Class (liftIO)
+import Criterion.Main (defaultMain, bgroup, bench, nfIO)
+import Criterion.Types (Benchmarkable(..))
+import qualified Data.Vector as V
```
blackgnezdo: Is TensorFlow.Core useful to shorten the list here?

tensorflow-ops/tests/FeedFetchBench.hs
```diff
@@ -0,0 +1,49 @@
...
+whnfSession :: TF.Session (a -> TF.Session b) -> a -> Benchmarkable
+whnfSession init x = Benchmarkable $ \m -> TF.runSession $ do
+    f <- init
+    -- Can't use replicateM because n is Int64.
+    let go n | n <= 0    = return ()
+             | otherwise = f x >>= liftIO . evaluate >> go (n-1)
+    go m
+
+-- | Benchmark feeding and fetching a vector.
+feedFetchBenchmark :: TF.Session (V.Vector Float -> TF.Session (V.Vector Float))
```
blackgnezdo: Is Vector strict enough to be fully constructed when evaluate runs this action? Maybe add a comment to this effect?

tensorflow-ops/tensorflow-ops.cabal
```diff
@@ -187,6 +187,14 @@ Test-Suite TypesTest
                , test-framework-quickcheck2
                , vector
 
+Benchmark FeedFetchBench
+  default-language: Haskell2010
+  type:       exitcode-stdio-1.0
+  main-is:    FeedFetchBench.hs
+  hs-source-dirs: tests
+  build-depends: base, criterion, tensorflow, tensorflow-ops, vector, transformers
```
blackgnezdo: Split over multiple lines and sort?

tensorflow-ops/tests/FeedFetchBench.hs
fkm3: Done.

tensorflow-ops/tests/FeedFetchBench.hs
fkm3: I think so, but I've switched to deepseq anyways to make sure. This makes the bench slower.

tensorflow/src/TensorFlow/Internal/FFI.hs
```diff
@@ -187,8 +188,12 @@ createTensorData t = do
     -- Read data.
     len <- fromIntegral <$> Raw.tensorByteSize t
     bytes <- castPtr <$> Raw.tensorData t :: IO (Ptr Word8)
-    -- TODO(fmayle): Don't copy the data.
-    v <- S.fromList <$> peekArray len bytes
+    -- Note: We would like to avoid the copy by creating an S.Vector directly
+    -- from 'bytes' and calling Raw.deleteTensor when it gets GC'd, but we can't
+    -- because the vector may outlive the tensorflow session and become invalid.
```
blackgnezdo: It sounds like what we'd like is to have SM.Vector with a custom destructor which we would hook up to call Raw.deleteTensor. Is this correct?

tensorflow/src/TensorFlow/Internal/FFI.hs
```diff
@@ -187,8 +188,12 @@ createTensorData t = do
...
+    mv <- SM.new len
```
blackgnezdo: Could you confirm the unints of SM.new are in bytes? There also appears to be some potential for an arithmetic overflow if fromIntegral at line 189 is not doing the right thing. Maybe externalize my safeFromIntegral?

blackgnezdo: Nevermind, TensorData is using Word8. The fromIntegral part is still suspect though as it seems to be :: CULong -> Int.

tensorflow/src/TensorFlow/Internal/FFI.hs
fkm3: Right. I had done that and was getting an error on process exit for the benchmark, which led me to think that the ffi tensors were being free'd on session deletion or something. I've looked into it more now and realize that they aren't associated with the session, so that wasn't the issue. I can't reproduce the error anymore... so, going with the implementation that doesn't copy at all.

tensorflow/src/TensorFlow/Internal/FFI.hs
fkm3: Done. I was looking for an existing version of your safeFromIntegral in the wild and found `toIntegralSized`, which does most of what we want and might be more efficient. Some of the type signatures got a bit uglier though. Let me know what you think.

PR #29, opened 2016-11-15
Uprev tensorflow.

blackgnezdo:
Includes temporary blacklisting for a couple of ops that will be
supported once my fix lands in the main tensorflow repo.


blackgnezdo:
We got a flakie test! HTTP downloads don't work sometimes. Since the rest of tests passed I'm overriding the trigger.

Issue #38, opened 2016-11-20
GHC 8


o1lo01ol1o:
Bumping the master stack.yaml to lts-7.3 unsurprisingly causes the build to break in OSX el capitan:

```
[...]
tensorflow-opgen-0.1.0.0: copy/register
Progress: 12/17
--  While building package tensorflow-0.1.0.0 using:
      /Users/.../.stack/setup-exe-cache/x86_64-osx/setup-Simple-Cabal-1.24.0.0-ghc-8.0.1 --builddir=.stack-work/dist/x86_64-osx/Cabal-1.24.0.0 build lib:tensorflow --ghc-options " -ddump-hi -ddump-to-file"
    Process exited with code: ExitFailure 1
    Logs have been written to: /Users/.../tensorflow-haskell/.stack-work/logs/tensorflow-0.1.0.0.log

    Configuring tensorflow-0.1.0.0...
    Preprocessing library tensorflow-0.1.0.0...
    [ 4 of 12] Compiling TensorFlow.Types ( src/TensorFlow/Types.hs, .stack-work/dist/x86_64-osx/Cabal-1.24.0.0/build/TensorFlow/Types.o )
    
    /Users/.../tensorflow-haskell/tensorflow/src/TensorFlow/Types.hs:376:5: error:
        Malformed head of type or class declaration: as \\ b : bs
```

  Is there any notion of when GHC 8 might be supported?

blackgnezdo:
Do you have a particular requirement forcing you to use GHC 8?

We would like to retain 7.10 compatibility for some time. If you can suggest a fix making both versions work, we'll probably take it.

o1lo01ol1o:
the datahaskell group would like to target ghc 8+ in working toward a consolidated datascience environment in haskell  (still a ways off, but nonetheless .  .  .).    

I'll try to take a look at the problematic bits if I get some time.  

Thanks for the reply.

judah:
Fixed by #97

PR #40, opened 2016-11-21
Fix the build on ghc-8.0.1 (#38).

judah:
Two issues:
- The definition of `\\` was missing parentheses.  It was probably a bug
  that this used to worked in ghc-7.10.
- Set -fconstraint-solver-iterations=0 to work around
  https://ghc.haskell.org/trac/ghc/ticket/12175.  It looks like we can
  trigger that bug when defining a significantly complicated op.  Specifically,
  our type shenanigans ("OneOf") along with lens setters (for OpDef) seem
  to confuse GHC.

Still TODO: automate testing of different ghc versions to prevent a regression.

### DIFF
--- a/tensorflow-core-ops/tensorflow-core-ops.cabal
+++ b/tensorflow-core-ops/tensorflow-core-ops.cabal
@@ -24,6 +24,10 @@ library
                 , lens-family
                 , text
   default-language:    Haskell2010
+  -- Work around https://ghc.haskell.org/trac/ghc/ticket/12175.
+  if impl(ghc >= 8) {
+    ghc-options: -fconstraint-solver-iterations=0
+  }
 
 source-repository head
   type:     git
--- a/tensorflow-nn/tensorflow-nn.cabal
+++ b/tensorflow-nn/tensorflow-nn.cabal
@@ -19,6 +19,10 @@ library
                 , tensorflow == 0.1.*
                 , tensorflow-ops == 0.1.*
   default-language:    Haskell2010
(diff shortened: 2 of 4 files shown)


tensorflow-core-ops/tensorflow-core-ops.cabal
```diff
@@ -24,6 +24,10 @@ library
                 , lens-family
                 , text
   default-language:    Haskell2010
+  -- Work around https://ghc.haskell.org/trac/ghc/ticket/12175.
```
blackgnezdo: Presumably this is only needed for < 8.0.2? Or am I misreading the bug report?

tensorflow-core-ops/tensorflow-core-ops.cabal
```diff
@@ -24,6 +24,10 @@ library
...
+  if impl(ghc >= 8) {
+    ghc-options: -fconstraint-solver-iterations=0
```
blackgnezdo: Any chance this can be centralized?

gunan:
If the build script has a parameter, It will be quite simple.
I would suggest modifying build script, and having the default as the latest compiler.
Then please ping me and I can quickly create a 2nd build job for you.

tensorflow-core-ops/tensorflow-core-ops.cabal
judah: It turns out that it's not completely fixed in ghc-8.0.2-rc1.  I filed a bug: https://ghc.haskell.org/trac/ghc/ticket/12885

PR #41, opened 2016-11-21
Make GHC 8 build succeed

nsuke:
I've adjusted one parenthesis and it seem to be working on GHC 8.0.1.
`-fconstraint-solver-iterations=0` flag is also needed to workaround following failure:
```
solveWanteds: too many iterations (limit = 4)
```
So concretely the build command looked like this:
```
stack --resolver lts-7.9 test --ghc-options -fconstraint-solver-iterations=0
```

FYI, due to commercialhaskell/stack#2712, currently stack's GHC does not work for some system (mine included), so it can be handy.

cc: #38


nsuke:
@judah nvm, you're already working on it.

Issue #42, opened 2016-11-22
Building .dylib on osx?


o1lo01ol1o:
The build instructions for osx produce a shared object `.so` file but this extension is not resolved by stack when trying to link the tensorflow project as a local dependency.  (It looks like the `.so` format may not play nicely with stack on osx.)  Is there a way to create a `.dylib` target and build that instead?  References [here](http://stackoverflow.com/questions/40713673/force-stack-to-look-for-so-files-instead-of-dylib-on-os-x) and [here](https://github.com/commercialhaskell/stack/issues/2798)

judah:
Sorry, it looks like I missed a step in the README.  It's enough to just symlink the `.so` file to the `.dylib`, though for some reason it only works if *both* files are present.

I've sent out #43 to update the build instructions.

PR #44, opened 2016-11-22
Update OS X instructions (#42) to not require a separate ".so" file.

judah:
The right approach is to run `install_name_tool` on the library after renaming
its extension from ".so" to ".dylib".


blackgnezdo:
Does install_name_tool update the file or change some kind of system database (like ldconfig)? If the file is in fact changed, maybe it should be changed *before* copied? Just to make sure the unadorned file is not left behind (much harder to diagnose than a missing file).

Another concern is maybe this whole magic belongs in bazel build rather than our install instructions?

Finally, how does one even trigger the problem? I imagine you verified things work as you were writing the instructions in the first place?

PR #49, opened 2016-12-12
Fix TensorData encode and decode for Bool

fkm3:


tensorflow-ops/tests/TypesTest.hs
```diff
@@ -37,6 +37,7 @@ import qualified Data.ByteString.Char8 as B8
 import qualified Data.Vector as V
 
 import qualified TensorFlow.ControlFlow as TF
+import qualified TensorFlow.GenOps.Core as TF (select)
```
blackgnezdo: Does it need to be both qualified and explicitly listed?

fkm3: I prefer qualified imports, so I'm stubbornly keeping it qualified. Have to explicitly list the imports or use `hiding` to avoid collisions with `TensorFlow.Ops`.

I think we should re-export all of the non-colliding gen ops from `TensorFlow.Ops` at some point.

PR #50, opened 2016-12-13
Support fetching storable vectors + use them in benchmark

fkm3:
In addition, you can now fetch TensorData directly. This might be useful in
scenarios where you feed the result of a computation back in, like RNN.

Before:

benchmarking feedFetch/4 byte
time                 83.31 μs   (81.88 μs .. 84.75 μs)
                     0.997 R²   (0.994 R² .. 0.998 R²)
mean                 87.32 μs   (86.06 μs .. 88.83 μs)
std dev              4.580 μs   (3.698 μs .. 5.567 μs)
variance introduced by outliers: 55% (severely inflated)

benchmarking feedFetch/4 KiB
time                 114.9 μs   (111.5 μs .. 118.2 μs)
                     0.996 R²   (0.994 R² .. 0.998 R²)
mean                 117.3 μs   (116.2 μs .. 118.6 μs)
std dev              3.877 μs   (3.058 μs .. 5.565 μs)
variance introduced by outliers: 31% (moderately inflated)

benchmarking feedFetch/4 MiB
time                 109.0 ms   (107.9 ms .. 110.7 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 108.6 ms   (108.2 ms .. 109.2 ms)
std dev              740.2 μs   (353.2 μs .. 1.186 ms)

After:

benchmarking feedFetch/4 byte
time                 82.92 μs   (80.55 μs .. 85.24 μs)
                     0.996 R²   (0.993 R² .. 0.998 R²)
mean                 83.58 μs   (82.34 μs .. 84.89 μs)
std dev              4.327 μs   (3.664 μs .. 5.375 μs)
variance introduced by outliers: 54% (severely inflated)

benchmarking feedFetch/4 KiB
time                 85.69 μs   (83.81 μs .. 87.30 μs)
                     0.997 R²   (0.996 R² .. 0.999 R²)
mean                 86.99 μs   (86.11 μs .. 88.15 μs)
std dev              3.608 μs   (2.854 μs .. 5.273 μs)
variance introduced by outliers: 43% (moderately inflated)

benchmarking feedFetch/4 MiB
time                 1.582 ms   (1.509 ms .. 1.677 ms)
                     0.970 R²   (0.936 R² .. 0.993 R²)
mean                 1.645 ms   (1.554 ms .. 1.981 ms)
std dev              490.6 μs   (138.9 μs .. 1.067 ms)
variance introduced by outliers: 97% (severely inflated)

### DIFF
--- a/tensorflow/src/TensorFlow/Nodes.hs
+++ b/tensorflow/src/TensorFlow/Nodes.hs
@@ -12,8 +12,8 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -24,7 +24,6 @@ import Control.Applicative (liftA2, liftA3)
 import Data.Map.Strict (Map)
 import Data.Monoid ((<>))
 import Data.Set (Set)
-import Data.String (IsString)
 import Data.Text (Text)
 import Lens.Family2 ((^.))
 import qualified Data.Map.Strict as Map
@@ -101,18 +100,12 @@ instance a ~ () => Fetchable ControlNode a where
 instance Nodes (Tensor v a) where
     getNodes t = Set.singleton <$> getOrAddOp (t ^. tensorOutput . outputOp)
 
-fetchTensorList :: TensorType a => Tensor v a -> Build (Fetch (Shape, [a]))
-fetchTensorList t = fmap (fmap V.toList) <$> fetchTensorVector t
-
 fetchTensorVector :: forall a v . TensorType a
-                  => Tensor v a -> Build (Fetch (Shape, V.Vector a))
+                  => Tensor v a -> Build (Fetch (TensorData a))
 fetchTensorVector (Tensor _ o) = do
     outputName <- renderOutput o
     return $ Fetch (Set.singleton outputName) $ \tensors ->
         let tensorData = tensors Map.! outputName
-            shape = Shape $ FFI.tensorDataDimensions tensorData
-            vec = decodeTensorData $ TensorData tensorData
-
             expectedType = tensorType (undefined :: a)
             actualType = FFI.tensorDataType tensorData
             badTypeError = error $ "Bad tensor type: expected "
@@ -121,21 +114,12 @@ fetchTensorVector (Tensor _ o) = do
                                    ++ show actualType
         in if expectedType /= actualType
                then badTypeError
-               else (shape, vec)
+               else TensorData tensorData
 
 -- The constraint "a ~ a'" means that the input/output of fetch can constrain
 -- the TensorType of each other.
-instance (TensorType a, a ~ a') => Fetchable (Tensor v a) (V.Vector a') where
-    getFetch t = fmap snd <$> fetchTensorVector t
+instance (TensorType a, a ~ a') => Fetchable (Tensor v a) (TensorData a') where
+    getFetch = fetchTensorVector
 
-newtype Scalar a = Scalar {unScalar :: a}
-    deriving (Show, Eq, Ord, Num, Fractional, Floating, Real, RealFloat,
-              RealFrac, IsString)
-
-instance (TensorType a, a ~ a') => Fetchable (Tensor v a) (Scalar a') where
-    getFetch t = fmap (Scalar . headFromSingleton . snd) <$> fetchTensorList t
-      where
-        headFromSingleton [x] = x
-        headFromSingleton xs
-            = error $ "Unable to extract singleton from tensor of length "
-                          ++ show (length xs)
+instance (TensorType a, TensorDataType s a, a ~ a') => Fetchable (Tensor v a) (s a') where
+    getFetch t = fmap decodeTensorData <$> fetchTensorVector t
--- a/tensorflow/src/TensorFlow/Nodes.hs
+++ b/tensorflow/src/TensorFlow/Nodes.hs
@@ -28,7 +28,6 @@ import Data.Text (Text)
 import Lens.Family2 ((^.))
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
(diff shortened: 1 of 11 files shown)


fkm3:
I couldn't find a way to deal with overlapping instances, so I just listed all of the instances manually. Tensorflow only supports a small set of element types, so this seems reasonable to me. Let me know what you think.

fkm3:
Half baked idea: Instead of

```haskell
class TensorDataType s a where
    ...

instance TensorDataType V.Vector Float
```

we could make it

```haskell
class TensorDataType a where
    ...

instance TensorDataType (V.Vector Float)
```

then, we could drop `Scalar` and allow scalars to be fetched directly

```haskell
instance TensorDataType Float
```

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -163,40 +133,173 @@ instance TensorType Word8 where
...
+
+instance TensorDataType V.Vector Int16 where
+    decodeTensorData = S.convert . simpleDecode
+    encodeTensorData x = simpleEncode x . S.convert
+
+instance TensorDataType S.Vector Int32 where
+    decodeTensorData = simpleDecode
+    encodeTensorData = simpleEncode
+
+instance TensorDataType V.Vector Int32 where
```
judah: Does this work to remove some of the redundancy?

    instance TensorDataType S.Vector a => TensorDataType V.Vector a where
        decodeTensorData = S.convert . decodeTensorData
        encodeTensorData x = simpleEncode x . S.convert

(maybe with some extra type annotations as in the Bool instance)

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -95,66 +100,31 @@ import Proto.Tensorflow.Core.Framework.Types (DataType(..))
 import TensorFlow.Internal.VarInt (getVarInt, putVarInt)
 import qualified TensorFlow.Internal.FFI as FFI
 
--- | Data about a tensor that is encoded for the TensorFlow APIs.
-newtype TensorData a = TensorData { unTensorData :: FFI.TensorData }
-
 -- | The class of scalar types supported by tensorflow.
-class TensorType a where
+class TensorDataType V.Vector a => TensorType a where
```
judah: This particular superclass constraint seems a little odd since I think for most types users will want to use S.Vector rather than V.Vector...does anything get more complicated if you just remove it?  (Since otherwise TensorType is usually about graph construction, not fetching.)

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -163,40 +133,173 @@ instance TensorType Word8 where
...
+instance TensorType (Complex Double) where
+    tensorType _ = DT_COMPLEX128
+    tensorRefType _ = DT_COMPLEX128
+    tensorVal = error "TODO (Complex Double)"
+
+
+-- | Tensor data with the correct memory layout for tensorflow.
+newtype TensorData a = TensorData { unTensorData :: FFI.TensorData }
+
+-- | Types that can be converted to and from 'TensorData'
```
judah: Period at the end of the sentence.

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -163,40 +133,173 @@ instance TensorType Word8 where
...
+--
+-- 'S.Vector' is the most effecient to encode/decode for most element types.
```
judah: 'efficient'

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -163,40 +133,173 @@ instance TensorType Word8 where
...
+class TensorDataType s a where
```
judah: With respect to your comment about removing the need for Scalar, maybe we could use type families:

    class TensorDataType a where
        type TensorDataTypeElem a
        decodeTensorData :: TensorData (TensorDataTypeElem a) -> a

etc.

Not necessary for this CL, though.

tensorflow/src/TensorFlow/Nodes.hs
```diff
@@ -101,18 +100,12 @@ instance a ~ () => Fetchable ControlNode a where
 instance Nodes (Tensor v a) where
     getNodes t = Set.singleton <$> getOrAddOp (t ^. tensorOutput . outputOp)
 
-fetchTensorList :: TensorType a => Tensor v a -> Build (Fetch (Shape, [a]))
-fetchTensorList t = fmap (fmap V.toList) <$> fetchTensorVector t
-
 fetchTensorVector :: forall a v . TensorType a
-                  => Tensor v a -> Build (Fetch (Shape, V.Vector a))
```
judah: If you're removing this instance, is there another way to access the Shape of a fetched Tensor?

tensorflow/src/TensorFlow/Types.hs
judah: Ah, I saw you mentioned a concern about overlapping instances.  To be honest this seems like a reasonable use case to mark the more general instance as `{-# OVERLAPPABLE #-}`.  But up to you.

tensorflow/src/TensorFlow/Types.hs
fkm3: It just reduced the code churn in our tests, but on second thought that isn't worth it.

I've made `TensorType` a superclass of `TensorDataType` instead, but that is just for convenience. Let me know what you think.

tensorflow/src/TensorFlow/Types.hs
fkm3: Done.

tensorflow/src/TensorFlow/Types.hs
fkm3: Done.

tensorflow/src/TensorFlow/Types.hs
fkm3: May give this a try in a separate PR.

tensorflow/src/TensorFlow/Types.hs
fkm3: Done.

tensorflow/src/TensorFlow/Nodes.hs
fkm3: This isn't an instance. I was thinking of `fetchTensorVector` as an internal function and you can still get the shape by using `TensorData`. Adding instance for `(Shape, ...)` sounds like a good idea though.

PR #53, opened 2016-12-22
adds `initializedValue`

silky:
as per https://github.com/blackgnezdo/tensorflow-haskell/commit/3f50193954f50dbbb3f773d8f719c887a66825a7 this attempts to introduce the `initialized_value` function.

do we want this? does this make sense?

### DIFF
--- a/tensorflow-ops/src/TensorFlow/Ops.hs
+++ b/tensorflow-ops/src/TensorFlow/Ops.hs
@@ -68,6 +68,7 @@ module TensorFlow.Ops
     , constant
     , CoreOps.equal
     , expandDims
+    , initializedValue
     , initializedVariable
     , zeroInitializedVariable
     , CoreOps.fill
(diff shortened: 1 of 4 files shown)


tensorflow-ops/src/TensorFlow/Ops.hs
```diff
@@ -161,6 +164,18 @@ placeholder shape' =
...
+
+-- | Construct a tensor whose value is the initialized value of the given
+-- tensor.
+initializedValue :: forall a v.  TensorType a 
+    => Tensor v a 
+    -> Build (Tensor v a)
+initializedValue t = do
+    ns <- use initializationNodes
+    -- Make this tensor depend on the initializers of the other.
+    withNodeDependencies (fromList ns) (render t)
```
blackgnezdo: I'm certainly impressed by how small the change is :)

On the other hand, this is a really big hammer which will impede concurrency (and result in O(n^2) graph size). This effect may also kick in unexpectedly when somebody refactors their code and a variable that was created with few `initializationNodes` suddenly get moved to a later point when a lot of variables are created prior to it.

silky: Okay, fair comments.

PR #58, opened 2016-12-30
Updated TensorFlow.

blackgnezdo:
Needed to list a newly added proto in cabal file.

### DIFF
--- a/ci_build/outer_launch_tests.sh
+++ b/ci_build/outer_launch_tests.sh
@@ -8,6 +8,5 @@ STACK_RESOLVER=${STACK_RESOLVER:-lts-6.2}
 IMAGE_NAME=tensorflow/haskell/ci_build:$STACK_RESOLVER
 
 git submodule update
-docker pull tensorflow/tensorflow:nightly-devel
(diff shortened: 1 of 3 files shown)


judah:
LGTM, but any idea why the CI build is failing?

ci_build/outer_launch_tests.sh
```diff
@@ -8,5 +8,6 @@ STACK_RESOLVER=${STACK_RESOLVER:-lts-6.2}
 IMAGE_NAME=tensorflow/haskell/ci_build:$STACK_RESOLVER
 
 git submodule update
+docker pull tensorflow/tensorflow:nightly-devel
```
judah: Not necessary in this PR, but: what do you think about making the dockerfiles more specific than "nightly" so they correspond more stably to the submodule's version?  For example, picking a release like `tensorflow/tensorflow:0.12.1-devel`:
https://hub.docker.com/r/tensorflow/tensorflow/tags/

blackgnezdo: I think 0.12.1-devel is not compatible with recent TF. We'll probably have to wait until newer tag before pinning it.

PR #60, opened 2017-01-14
Add example to README + make haddock link more prominent

fkm3:
Let me know if you have any better ideas for a "just show me some code" sort of example.

### DIFF
--- a/tensorflow-ops/examples/Regression.hs
+++ b/tensorflow-ops/examples/Regression.hs
@@ -1,0 +1,46 @@
+-- | Simple linear regression example for the README.
+
+import Control.Monad (replicateM, replicateM_, zipWithM)
+import System.Random (randomIO)
+import Text.Printf (printf)
+
+import qualified TensorFlow.Core as TF
+import qualified TensorFlow.GenOps.Core as TF
+import qualified TensorFlow.Gradient as TF
+import qualified TensorFlow.Ops as TF
+
+main :: IO ()
+main = do
+    -- Generate data where `y = x*3 + 8`.
+    xData <- replicateM 100 randomIO
+    let yData = [x*3 + 8 | x <- xData]
+    -- Fit linear regression model.
+    (w, b) <- fit xData yData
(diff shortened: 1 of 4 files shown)


tensorflow-ops/tensorflow-ops.cabal
```diff
@@ -30,6 +30,18 @@ library
                 , text
   default-language:    Haskell2010
 
+
```
judah: Surround the executable block with a single blank line, similar to the rest of this file.

tensorflow-ops/examples/Regression.hs
```diff
@@ -0,0 +1,46 @@
...
+    let yHat = (x `TF.mul` w) `TF.add` b
+        loss = TF.square (yHat `TF.sub` y)
+    -- Optimize with gradient descent.
+    trainStep <- TF.build (gradientDescent 0.001 loss [w, b])
+    replicateM_ 1000 (TF.run trainStep)
+    -- Return the learned parameters.
+    (TF.Scalar w', TF.Scalar b') <- TF.run (w, b)
+    return (w', b')
+
+gradientDescent :: Float
```
judah: Consider including `gradientDescent` in the README example to make it self-contained, since it's only a few more lines?  Up to you.

blackgnezdo: Unless we compile the code in README.md it will grow stale. Having the complete compiled (better yet - tested) sample makes it more likely to work.

tensorflow-ops/tensorflow-ops.cabal

tensorflow-ops/examples/Regression.hs
fkm3: It doesn't look like there is a way to include a source file into the markdown, so as to force them to be in sync. I've included the full source in the example now + turned it into a test.

fkm3:
FYI, you can view the rendered markdown here: https://github.com/fkm3/tensorflow-haskell/tree/readme_example

PR #69, opened 2017-02-08
Uprev to TF 1.0rc1.

blackgnezdo:
* Download protoc and libtensorflow instead of running bazel.


judah:
The description mentions hiding ResourceHandle ops; I didn't see such a change in these commits/

blackgnezdo:
Removed the ResourceHandle note. It was out of date.

PR #70, opened 2017-02-09
Add tensorflow-records and tensorflow-records-conduit package

fkm3:
Original author: @awpr

### DIFF
--- a/tensorflow-records/src/TensorFlow/CRC32C.hs
+++ b/tensorflow-records/src/TensorFlow/CRC32C.hs
@@ -1,0 +1,55 @@
+-- Copyright 2016 TensorFlow authors.
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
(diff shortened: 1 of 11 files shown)


fkm3:
@awpr, let me know if you'd rather take this over. If not, I'll try to edit you in as the commit author before merging.

tensorflow-records/src/TensorFlow/CRC32C.hs
```diff
@@ -0,0 +1,55 @@
...
+import qualified Data.ByteString as B
+import qualified Data.ByteString.Lazy as BL
+import Data.Digest.CRC32C (crc32c_update)
+import Data.List (foldl')
+import Data.Word (Word32)
+
+-- | Compute the CRC32C checksum of the concatenation of the bytes checksummed
+-- by the given CRC32C value and the bytes in the given ByteString.
+extend :: Word32 -> B.ByteString -> Word32
+extend = crc32c_update
```
fkm3: Changed this to use Data.Digest.CRC32C instead of FFI.

tensorflow-records/src/TensorFlow/Records.hs
```diff
@@ -0,0 +1,131 @@
...
+-- | Put a record length and its checksum.
+putTFRecordLength :: Word64 -> Put
+putTFRecordLength x =
+  let put = putWord64le x
+      len = 8
+      crc = CRC.mask $ unsafePerformIO $
+          -- Serialized Word64 is always 8 bytes, so we can go fast by using
+          -- alloca.
+          unsafeWithFixedWidthBuilder len (execPut put) $
+              \ptr -> CRC.extend 0 <$> unsafePackCStringLen (castPtr ptr, len)
```
fkm3: Changed this to use `extend` instead of `nativeExtend`.

awpr:
Thanks for taking the time to externalize this!  I'd like to move a few things around now that it's using an external CRC32C implementation, and I've already figured out how to get separate author/committer in git, so I can do those two things and get the history just right at the same time.

PR #74, opened 2017-02-13
Add support for logging to tensorboard

fkm3:
Based on @gnezdo's internal version with some differences:

* Uses a pure haskell implementation of EventWriter instead of FFI.
* Special `buildAnd*` functions were dropped in favor of using
  `mergeAllSummaries :: Build SummaryTensor` with the normal
  `build` function.


tensorflow-ops/src/TensorFlow/Logging.hs
```diff
@@ -0,0 +1,158 @@
...
+-- > -- Write summaries to an EventWriter.
+-- > train = TF.withEventWriter "/path/to/logs" $ \eventWriter -> do
+-- >     summaryTensor <- TF.build TF.allSummaries
+-- >     forM_ [1..] $ \step -> do
+-- >         if (step % 100 == 0)
+-- >             then do
+-- >                 ((), summaryBytes) <- TF.run (trainStep, summaryTensor)
+-- >                 let summary = decodeMessageOrDie (TF.unScalar summaryBytes)
+-- >                 TF.logSummary eventWriter step summary
+-- >             else TF.run_ trainStep
```
fkm3: The interface feels a bit awkward. I wonder if we could add `SummaryTensor` as a separate type instead of an alias of `Tensor Value ByteString` and then handle the decoding automatically.

tensorflow-ops/tensorflow-ops.cabal
```diff
@@ -16,18 +16,31 @@ library
   exposed-modules: TensorFlow.Gradient
                  , TensorFlow.Ops
                  , TensorFlow.EmbeddingOps
+                 , TensorFlow.Logging
   build-depends:  proto-lens == 0.1.*
                 , base >= 4.7 && < 5
                 , bytestring
-                , fgl
-                , mtl
+                , conduit
```
blackgnezdo: As you noted, having heavy dependencies in the core will slow things down. I vote for introducing tensorflow-logging, unless you believe everything will have logging.

tensorflow-logging/src/TensorFlow/Logging.hs
```diff
@@ -0,0 +1,158 @@
...
+import qualified TensorFlow.GenOps.Core as CoreOps
+
+-- | Handle for logging TensorBoard events safely from multiple threads.
+data EventWriter = EventWriter (TBMQueue Event) (MVar ())
+
+-- | Writes Event protocol buffers to event files.
+withEventWriter ::
+    (MonadIO m, MonadMask m)
+    => FilePath
+    -- ^ logdir. Directory where event file will be written.
```
blackgnezdo: Harmonize logdir vs filePrefix and newEventWriter logdir?

tensorflow-logging/src/TensorFlow/Logging.hs
```diff
@@ -0,0 +1,158 @@
...
+newEventWriter :: FilePath -> IO EventWriter
+newEventWriter logdir = do
+    createDirectoryIfMissing True logdir
+    -- Create unique filename from prefix.
+    t <- doubleWallTime
+    hostname <- getHostName
+    let filename = printf (logdir </> "events.out.tfevents.%010d.%s")
+                          (truncate t :: Integer) hostname
+    -- Asynchronously consume events from a queue.
+    q <- newTBMQueueIO 1024
```
blackgnezdo: Document why 1024?

tensorflow-logging/tests/LoggingTest.hs
```diff
@@ -0,0 +1,63 @@
...
+    withSystemTempDirectory "event_writer_logs" $ \dir -> do
+        assertEqual "No file before" [] =<< listDirectory dir
+        let expected = [ def & step .~ 10
+                       , def & step .~ 222
+                       , def & step .~ 8
+                       ]
+        withEventWriter dir $ \eventWriter ->
+            mapM_ (logEvent eventWriter) expected
+        files <- listDirectory dir
+        assertBool "One file exists after" (length files == 1)
```
blackgnezdo: This should give a more informative error message:
assertEqual 1 (length files)

tensorflow-logging/tests/LoggingTest.hs
```diff
@@ -0,0 +1,63 @@
...
+        records <- runResourceT $ Conduit.runConduit $
+            sourceTFRecords (dir </> head files) =$= Conduit.consume
+        let (header:body) = decodeMessageOrDie . BL.toStrict <$> records
```
blackgnezdo: Assert that length of records > 0? Just to save debugging grief.

tensorflow-logging/src/TensorFlow/Logging.hs
fkm3: Done.

tensorflow-logging/src/TensorFlow/Logging.hs
fkm3: Done.

tensorflow-logging/tests/LoggingTest.hs
fkm3: Done.

tensorflow-logging/tests/LoggingTest.hs
fkm3: Done.

tensorflow-ops/src/TensorFlow/Logging.hs
```diff
@@ -0,0 +1,158 @@
...
+import qualified TensorFlow.GenOps.Core as CoreOps
+
+-- | Handle for logging TensorBoard events safely from multiple threads.
+data EventWriter = EventWriter (TBMQueue Event) (MVar ())
+
+-- | Writes Event protocol buffers to event files.
+withEventWriter ::
+    (MonadIO m, MonadMask m)
+    => FilePath
+    -- ^ logdir. Directory where event file will be written.
```
blackgnezdo: Let's specify explicitly the limitation (which was a reason for me to use FFI internally):
-- ^ logdir. Local filesystem directory where event file will be written.

fkm3: Done. I hadn't thought about that, but it shouldn't be too difficult to generalize later.

PR #79, opened 2017-02-24
Support lists of tensors in ops.

judah:
Adds a new type `ListOf` which wraps a heterogeneous list; for example,
`ListOf (Tensor Value) '[Int32, Float]` represents a list of two
elements: a tensor of int32s and a tensor of floats.

Also changes the `Queue2` type (which suppored pairs of tensors) to
`Queue` (which supports arbitrary lists).

### DIFF
--- a/tensorflow-opgen/src/TensorFlow/OpGen.hs
+++ b/tensorflow-opgen/src/TensorFlow/OpGen.hs
@@ -314,17 +314,18 @@ tensorArg p = case parsedArgCase p of
     ResourceArg -> "ResourceHandle"
     SimpleArg { argType = t, argCaseKind = k } -> tensorType t k
     ListArg { argType = t, argCaseKind = k } -> brackets $ tensorType t k
-    MixedListArg {argTypeAttr = t} -> "TensorList Value" <+> renderHaskellName t
+    MixedListArg {argTypeAttr = t, argCaseKind = k}
+        -> "TensorList" <+> kind k <+> renderHaskellName t
   where
+    kind k = case k of
+                ArgTensorRef -> "Ref"
+                ArgTensorValue -> "Value"
+                ArgTensorEither v' -> strictText v'
     tensorType t k = let
-        v = case k of
-                ArgTensorRef -> "Tensor Ref"
-                ArgTensorValue -> "Tensor Value"
-                ArgTensorEither v' -> "Tensor" <+> strictText v'
         a = case t of
                 ArgTypeFixed dt -> strictText $ dtTypeToHaskell dt
                 ArgTypeAttr n -> renderHaskellName n
-        in v <+> a
+        in "Tensor" <+> kind k <+> a
 
 attrComment :: Attr a -> Doc
 attrComment a = argComment' (attrName a) (attrDescription a)
--- a/tensorflow-opgen/src/TensorFlow/OpGen.hs
+++ b/tensorflow-opgen/src/TensorFlow/OpGen.hs
@@ -147,6 +147,7 @@ imports = stack [
       "import Data.ByteString (ByteString)"
     , "import Data.Complex (Complex)"
     , "import Data.Int (Int8, Int16, Int32, Int64)"
+    , "import Data.Proxy (Proxy(Proxy))"
     , "import Data.Word (Word8, Word16)"
     , "import Lens.Family2 ((.~), (&))"
     , "import TensorFlow.Build"
@@ -229,9 +230,8 @@ functionBody pOp = buildFunction <+> parens (hang 0 (stack buildOpParts))
                             <- parsedOutputs pOp]
     buildOpParts =
         "opDef" <+> renderQuotedTFName (parsedOpName pOp) :
-        -- Renders tensor arguments.
-        [ "& opAttr" <+> renderQuotedTFName n <+>
-          ".~ tensorType (undefined ::" <+> renderHaskellName n <> ")"
+        -- Renders type parameter arguments.
+        [ "& opAttr" <+> renderQuotedTFName n <+> ".~" <+> inferredTypeExpr a
         | a <- inferredTypeAttrs pOp, let n = attrName a
         ] ++
         -- Renders mandatory attributes as function parameters.
@@ -244,6 +244,12 @@ functionBody pOp = buildFunction <+> parens (hang 0 (stack buildOpParts))
         ]
 
     tensorArgs = renderHaskellName . parsedArgName <$> parsedInputs pOp
+    inferredTypeExpr a
+        | typeParamIsList $ attrInfo a
+            = "fromTensorTypes (Proxy :: Proxy" <+> renderHaskellName (attrName a)
+                    <> ")"
+        | otherwise = "tensorType (undefined ::" <+> renderHaskellName (attrName a)
+                            <> ")"
 
 -- | Write a comment with the inputs/outputs/attributes in proto format, for
 -- debugging.
@@ -273,7 +279,7 @@ typeSig pOp = constraints
     typeParams = [strictText v | k <- parsedInputs pOp ++ parsedOutputs pOp,
                   Just (ArgTensorEither v) <- [argKind $ parsedArgCase k]]
                 ++ [renderHaskellName $ attrName n | n <- inferredTypeAttrs pOp]
-    classConstraints = tuple $ concatMap tensorArgConstraint
+    classConstraints = tuple $ map tensorArgConstraint
                     $ inferredTypeAttrs pOp
     signatureFold = folddoc (\x y -> x </> "->" <+> y)
     attrInput a = renderAttrType (attrInfo a) <+> hang 0 ("-- ^" <+> attrComment a)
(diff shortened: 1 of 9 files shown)


blackgnezdo:
Could you fix up the -Werror problem?

14:02:28     /tfhs/tensorflow/src/TensorFlow/Types.hs:407:5: Warning:
14:02:28         Pattern match(es) are non-exhaustive
14:02:28         In an equation for `==':
14:02:28             Patterns not matched:
14:02:28                 Nil (_ :| _)
14:02:28                 (_ :| _) Nil
14:02:28     
14:02:28     <no location info>: 
14:02:28     Failing due to -Werror.

https://ci.tensorflow.org/job/tensorflow-haskell-pull-requests/91/console

tensorflow-opgen/src/TensorFlow/OpGen/ParsedOp.hs
```diff
@@ -104,6 +103,13 @@ data AttrBaseType = AttrBytes | AttrInt64 | AttrFloat | AttrBool
                 | AttrType | AttrShape | AttrTensor
                 deriving Eq
 
+data TypeParam = TypeParam
+    { typeParamIsList :: Bool
+    , typeParamRestrictions :: Maybe [DataType]
+        -- ^ The list of allowed types (see: TensorFlow.Types.OneOf).
+        -- If 'Nothing', then any type is acceptable.
```
blackgnezdo: What does Just [] indicate? Does it make sense to drop Maybe and use [] as anything-goes value? Alternatively, Maybe NonEmpty would work.

tensorflow-queue/src/TensorFlow/Queue.hs
```diff
@@ -12,67 +12,65 @@
...
     (handle q)
 
 -- | Retrieves the values from the queue.
-dequeue :: forall a b . (TensorType a, TensorType b)
-           => Queue2 a b
-           -> Build (Tensor Ref a, Tensor Ref b)
+dequeue :: forall as . TensorTypes as
+           => Queue as
+           -> Build (TensorList Ref as)
            -- ^ Dequeued tensors. They are paired in a sense
```
blackgnezdo: paired -> coupled?

tensorflow/src/TensorFlow/Nodes.hs
```diff
@@ -96,6 +99,21 @@ instance Nodes ControlNode where
 instance a ~ () => Fetchable ControlNode a where
     getFetch _ = return $ pure ()
 
+-- hm, annoying but doable.
+-- unfortunately, pulling out requires a "Identity" everywhere...bleh
```
blackgnezdo: The confusing part is that it's not exactly everywhere. You can try to hide it a bit by composing :| with Identity into a different operator.

tensorflow-queue/tests/QueueTest.hs
```diff
@@ -39,42 +41,45 @@ import qualified Data.ByteString as BS
...
+    liftIO $ expected @=? y
 
 testAsync :: Test
 testAsync = testCase "testAsync" $ runSession $ do
     (deq, pump) <- build $ do
-        q :: Queue2 Int64 BS.ByteString <- makeQueue2 2 ""
+        q :: Queue [Int64, BS.ByteString] <- makeQueue 2 ""
         (,) <$> dequeue q
-            <*> enqueue q 10 (scalar "Async")
+            <*> enqueue q (10 :| scalar "Async" :| Nil)
```
blackgnezdo: It would be really good to explain why this 10 doesn't need any Identity whereas Scalar 10 below does.

tensorflow-queue/src/TensorFlow/Queue.hs
judah: Done.

tensorflow-opgen/src/TensorFlow/OpGen/ParsedOp.hs
judah: Switched to NonEmpty.

tensorflow-opgen/src/TensorFlow/OpGen.hs
```diff
@@ -244,6 +244,12 @@ functionBody pOp = buildFunction <+> parens (hang 0 (stack buildOpParts))
         ]
 
     tensorArgs = renderHaskellName . parsedArgName <$> parsedInputs pOp
+    inferredTypeExpr a
+        | typeParamIsList $ attrInfo a
```
judah: Done (and used elsewhere in this file).

judah:
I fixed the `-Werror` issue by adding a case that's #ifdef'd away for ghc-8, since that compiler is smart enough to know that the case isn't needed due to how the GADT types interact.

tensorflow/src/TensorFlow/BuildOp.hs
```diff
@@ -28,11 +30,13 @@ import Control.Monad (replicateM)
 import Control.Monad.Reader (ReaderT, runReaderT, ask)
 import Control.Monad.State.Strict (State, runState, get, put)
 import Data.Int (Int64)
+import Data.Proxy (Proxy(..))
```
blackgnezdo: Looks like it is unneeded and failing with -Werror.

tensorflow/src/TensorFlow/Nodes.hs
judah: Added (|:|) which is the version with Identity.

I feel it should be possible to simplify things, but haven't cracked it yet (and it may require ghc-8).  If I figure it out I'll send another PR.

tensorflow-queue/tests/QueueTest.hs
judah: Added a comment (in `testBasic`).

tensorflow/src/TensorFlow/BuildOp.hs
judah: Done.

tensorflow/src/TensorFlow/Nodes.hs
```diff
@@ -96,6 +99,21 @@ instance Nodes ControlNode where
...
+
+instance (Nodes (f a), Nodes (ListOf f as)) => Nodes (ListOf f (a ': as)) where
+    getNodes (x :| xs) = liftA2 Set.union (getNodes x) (getNodes xs)
+
+instance l ~ List '[] => Fetchable (ListOf f '[]) l where
+    getFetch _ = return $ pure Nil
+
+instance (Fetchable (f t) a, Fetchable (ListOf f ts) (List as), i ~ Identity)
+    => Fetchable (ListOf f (t ': ts)) (ListOf i (a ': as)) where
+    getFetch (x :| xs) = liftA2 (\y ys -> Identity y :| ys) <$> getFetch x <*> getFetch xs
```
blackgnezdo: |:| ?

tensorflow/src/TensorFlow/Tensor.hs
```diff
@@ -12,20 +12,31 @@
...
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
```
blackgnezdo: Pulling out big guns here? Maybe put a comment where it's needed?

tensorflow/src/TensorFlow/Nodes.hs
judah: Done.

tensorflow/src/TensorFlow/Tensor.hs
judah: Removed UndecidableInstances since it's not actually needed in this module.

awpr:
Heterogeneous list approach looks good.  I think it's also achievable by moving the constraints into the list type itself (with existential types), so the argument type for type-lists would be something like "SomeListOf (Tensor v) allowedTypes" but I don't think that would be any nicer than this.


I wonder if it's possible to have a typeclass method 'tensorList' that takes arbitrarily many arguments (selecting an (->) instance each time an argument is applied) and then gives a ListOf (Tensor v) ts when it's used in a context that expects ListOf.  (Essentially it would be 'hBuild' from the HList package).

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -376,6 +390,44 @@ instance Attribute [DataType] where
 instance Attribute [Int64] where
     attrLens = list . i
 
+-- | A heterogeneous list type.
+data ListOf f as where
+    Nil :: ListOf f '[]
+    (:|) :: f a -> ListOf f as -> ListOf f (a ': as)
```
judah: Done; renamed to (:/) and (/:/).

judah:
I think the existential types would make it more difficult to express operations like

    enqueue :: TensorTypes as => Queue as -> TensorList v as -> Build ControlNode

where we want the care about the two arguments being consistent, i.e., referring to the same list of types.

The `tensorList` class method is a nice idea.  I guess we'd need two variants; one for `ListOf` and one for `List` (similar to `:/` and `/:/`).  Due to that complexity, I prefer to punt for now.  Though maybe the class would let us avoid type operators altogether, like HList does.  (I'm not sure whether you can do something similar to overload pattern matches.)

awpr:
Sure, the typeclassy hlist-builder is just a syntactic convenience.  I agree the existential approach looks worse after all.  LGTM

PR #80, opened 2017-03-06
Added installation script for OSX dependencies under tools/

avctrh:
Added a script under tools/ to carry out OSX dependency installation steps https://github.com/tensorflow/haskell#build-on-mac-os-x

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,9 @@ There is also a demo application:
 
 ## Build on Mac OS X
 
-The following instructions were verified with Mac OS X El Capitan.
+The following instructions were verified with Mac OS X El Capitan. Alternatively
+run the [install_osx_dependencies.sh](./tools/install_osx_dependencies.sh)
+script in the `tools/` directory.
 
 - Install dependencies via [Homebrew](http://brew.sh):
 
--- a/README.md
+++ b/README.md
@@ -96,4 +96,4 @@ library on your machine under `/usr/local`.
 
 After running the script to install system dependencies, build the project with stack: 
 
-        stack test
+    stack test
--- a/README.md
+++ b/README.md
@@ -89,27 +89,7 @@ There is also a demo application:
 
 ## Build on Mac OS X
 
-The following instructions were verified with Mac OS X El Capitan. Alternatively
-run the [install_osx_dependencies.sh](./tools/install_osx_dependencies.sh)
-script in the `tools/` directory.
-
-- Install dependencies via [Homebrew](http://brew.sh):
-
-        brew install protobuf
-        brew install snappy
-
-- Install the TensorFlow library on your machine:
-
-        curl https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.0.0.tar.gz > libtensorflow.tar.gz
-        tar zxf libtensorflow.tar.gz -C /usr/local
-        mv /usr/local/lib/libtensorflow.so /usr/local/lib/libtensorflow.dylib
-        install_name_tool -id libtensorflow.dylib /usr/local/lib/libtensorflow.dylib
-        rm libtensorflow.tar.gz
-        
-- Install submodule dependencies:
-
-        git submodule update --init --recursive
-
-- Run stack:
-
-        stack test
+Run the [install_osx_dependencies.sh](./tools/install_osx_dependencies.sh)
+script in the `tools/` directory. The script installs dependencies
+via [Homebrew](http://brew.sh) and then downloads and installs the TensorFlow
+library on your machine under `/usr/local`.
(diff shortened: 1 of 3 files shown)


avctrh:
I signed the CLA.

blackgnezdo:
+judahjacobson, do you prefer a script or keep the textual instructions? I think we should pick one way we know works rather than two.

+avctrh thanks for signing the CLA!

tools/install_osx_dependencies.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
```
blackgnezdo: Please use
set -eu

to avoid the script plowing on after failures.

tools/install_osx_dependencies.sh
```diff
@@ -0,0 +1,40 @@
...
+echo "Extracting and copying libtensorflow..."
+tar zxf libtensorflow.tar.gz -C /usr/local
+mv /usr/local/lib/libtensorflow.so /usr/local/lib/libtensorflow.dylib
+
+echo -n "sudo password (required by install_name_tool) : "
+read -s spss
+echo ""
+echo $spss | sudo -S install_name_tool -id libtensorflow.dylib /usr/local/lib/libtensorflow.dylib
+rm libtensorflow.tar.gz
```
blackgnezdo: Should move above immediately after the archive is extracted.

tools/install_osx_dependencies.sh
```diff
@@ -0,0 +1,40 @@
...
+
+echo "Downloading libtensorflow..."
+curl https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.0.0.tar.gz > libtensorflow.tar.gz
+
+echo "Extracting and copying libtensorflow..."
+tar zxf libtensorflow.tar.gz -C /usr/local
+mv /usr/local/lib/libtensorflow.so /usr/local/lib/libtensorflow.dylib
+
+echo -n "sudo password (required by install_name_tool) : "
+read -s spss
```
blackgnezdo: From security point of view, it's best for users to not enter passwords into untrusted scripts (especially provided by a third party). So, we should either ask the user to run the whole script with sudo or output a command line with sudo which the user should run on their own.

tools/get_mnist_webarchive.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+echo "Downloading mnist datasets from web.archive.org"
```
blackgnezdo: This is a reasonable this to make the default. Please create a separate PR which changes urlPrefix in tensorflow-mnist-input-data/Setup.hs to be web.archive.org. Then this file should be reverted.

avctrh:
Thanks for the review @blackgnezdo let me know if any changes are needed.

judah:
I actually just set up another MacBook to build tensorflow-haskell and stumbled over a couple of the steps myself.  So I agree with blackgnezdo@ that we should remove the manual steps in the README altogether and replacing them with an instruction to just run this script.

avctrh:
@judah glad to hear this is a helpful direction.

@blackgnezdo I looked into migrating to webarchive for mnist w/ a separate PR in mind. The links require an extra bit of redirect handling, so I would switch from Network.HTTP to Network.Browser (in the same HTTP package). If that's okay I'll go ahead and work on it.

Let me know if anything else is needed for this PR.

README.md
```diff
@@ -89,7 +89,9 @@ There is also a demo application:
 
 ## Build on Mac OS X
 
-The following instructions were verified with Mac OS X El Capitan.
+The following instructions were verified with Mac OS X El Capitan. Alternatively
```
judah: To clarify my previous comment: let's remove all the manual steps listed in this section, and instead only suggest running the `install_osx_dependencies.sh` script.  (Otherwise, I think, we're begging for one or the other to get out of sync in the long run.)

We can still give a summary of what the script does in this section; for example:

    The script installs dependencies via [Homebrew](http://brew.sh) and then
    downloads and installs the TensorFlow library on your machine under
     `/usr/local`.

avctrh: Okay - I've removed the steps from the README.

blackgnezdo:
test this please

blackgnezdo:
@avctrh, switching to a more advanced http agent is fine, especially if it's in the same package, i.e. no dependency bloat.

avctrh:
@blackgnezdo @judah switch to web.archive.org downloads is prepped:

https://github.com/avctrh/tensorflow-haskell/commit/151a4748686e88e0cf56ba8a965040c20d1f50a7

I'll wait for this PR to be resolved before opening another. Incidentally, looks like this mnist download issue is causing the current CI to fail.

avctrh:
@blackgnezdo not sure who that was directed to, but I've tested it locally at least.

CI tests aren't passing due to MNIST downloads mentioned above. I've initiated a separate PR for that.

avctrh:
Great to hear!

PR #84, opened 2017-03-19
Add versions of each op that take optional params as an extra arg.

judah:
Each op `foo :: ...` now has a corresponding `foo' :: OpParams -> ...`
which lets you set optional attributes.  `OpParams` is currently a type alias for
`OpDef -> OpDef`.  In the future we should consider more type safety, e.g.,
using type-level strings and OverloadedLabels for optional attributes.

I used it to replace a few manual `buildOp`s in our code with the codegenerated
ops, now that it's easier to set attributes.  I also removed `tensorAttr` and
`named` since it's now possible to set those op attributes directly.

Although this clutters up the API a bit, I think it's simpler than using type
classes to implement optional arguments (as in, for example, `Text.Printf`) --
especially in terms of type inference with the rest of the library.

### DIFF
--- a/tensorflow-ops/src/TensorFlow/Ops.hs
+++ b/tensorflow-ops/src/TensorFlow/Ops.hs
@@ -58,56 +58,99 @@
 
 module TensorFlow.Ops
     ( CoreOps.add
+    , CoreOps.add'
     , CoreOps.abs
+    , CoreOps.abs'
     , CoreOps.addN
+    , CoreOps.addN'
     , CoreOps.argMax
+    , CoreOps.argMax'
     , CoreOps.assign
+    , CoreOps.assign'
     , CoreOps.broadcastGradientArgs
+    , CoreOps.broadcastGradientArgs'
     , CoreOps.cast
+    , CoreOps.cast'
     , CoreOps.concat
+    , CoreOps.concat'
     , constant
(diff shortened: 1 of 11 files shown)


tensorflow-ops/src/TensorFlow/Ops.hs
```diff
@@ -151,37 +194,44 @@ instance ( TensorType a
...
-placeholder shape' =
-    build $ buildOp $ opDef "Placeholder"
-            & opAttr "dtype" .~ tensorType (undefined :: a)
-            & opAttr "shape" .~ shape'
+placeholder :: (MonadBuild m, TensorType a) => Shape -> m (Tensor Value a)
+placeholder = placeholder' id
+
+placeholder' :: (MonadBuild m, TensorType a) => OpParams -> Shape -> m (Tensor Value a)
+placeholder' params pShape
+    = render $ CoreOps.placeholder' (params . (opAttr "shape" .~ pShape))
```
fkm3: The new version of placeholder can get CSE'd. Not sure what a good solution is besides reverting or having list of opgen overrides.

judah: Good catch.  I've sent #86 to revert the change to this op.

PR #87, opened 2017-03-23
Add another test of CSE and feeds.

judah:
As a follow-up to #86, check that our CSE isn't too aggressive to prevent feeds
of pure ops with distinct names.


judah:
Jenkins, test this please.

PR #90, opened 2017-04-03
Update the Stack resolver lts-8.5

Lupino:


blackgnezdo:
@judah, I think we are still interested in retaining GHC 7.10 compatibility, what do you say?

@Lupino thanks for the PR, it's good to have the fixes necessary for transitioning to a newer set of dependencies all lined up. I'm just not sure we can merge this change without having CIs for both 7.10 and the newest hotness.

avctrh:
FWIW, my experience has been GHC 8+ is now around long enough that the kinks seem to be ironed out and most active projects have migrated. Not sure if you have there's a specific reason to target 7.10, but I don't see a compelling reason to stay behind.

blackgnezdo:
Jenkins test this please.

PR #97, opened 2017-04-11
Adapt to lts-8.6 and use proto-lens-0.2.0.1

jcberentsen:
Squashed previous pull request

### DIFF
--- a/tensorflow/src/TensorFlow/Session.hs
+++ b/tensorflow/src/TensorFlow/Session.hs
@@ -49,7 +49,7 @@ import Data.Set (Set)
 import Data.Text.Encoding (encodeUtf8)
 import Lens.Family2 (Lens', (^.), (&), (.~))
 import Lens.Family2.Unchecked (lens)
-import Proto.Tensorflow.Core.Framework.Graph (node)
+import Proto.Tensorflow.Core.Framework.Graph (GraphDef, node)
 import Proto.Tensorflow.Core.Protobuf.Config (ConfigProto)
 import TensorFlow.Build
 import TensorFlow.Nodes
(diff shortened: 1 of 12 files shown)


judah:
Thanks for the patch.  Can you please confirm that it builds warning-free with `stack clean; stack build --ghc-options='-Wall -Werror'`?

In particular, while testing out the proto-related changes just now I noticed that you'll need to fix a typo of mine:

    --- a/tensorflow/src/TensorFlow/Types.hs
    +++ b/tensorflow/src/TensorFlow/Types.hs
    @@ -410,7 +410,7 @@ instance All Eq (Map f as) => Eq (ListOf f as) where
         (x :/ xs) == (y :/ ys) = x == y && xs == ys
         -- Newer versions of GHC use the GADT to tell that the previous cases are
         -- exhaustive.
    -#if _GLASGOW_HASKELL__ < 800
    +#if __GLASGOW_HASKELL__ < 800
         _ == _ = False
     #endif

jcberentsen:
I did notice that warning at first build. I should have looked closer. I'll
fix.

Den tir. 11. apr. 2017, 10:24 skrev Judah Jacobson <notifications@github.com

tensorflow/src/TensorFlow/Session.hs
```diff
@@ -135,7 +135,7 @@ extend = do
     trace <- Session (asks tracer)
     nodesToExtend <- build flushNodeBuffer
     unless (null nodesToExtend) $ liftIO $ do
-        let graphDef = def & node .~ nodesToExtend
+        let graphDef = (def :: GraphDef) & node .~ nodesToExtend
```
judah: For future reference (no changes needed for this PR): `proto-lens-0.2.*` made this extra type annotation necessary.  However, we can avoid it by switching the tensorflow packages away from `lens-family`.  Any of `lens-family-core`, `microlens` or `lens` would be sufficient.

Some notes on this topic are here:
https://github.com/google/proto-lens/blob/master/proto-lens-tests/tests/labels_test.hs#L40

judah:
Jenkins, test this please.

judah:
LGTM

PR #98, opened 2017-04-15
Add resource-based variable ops.

### LINKED ISSUE #92: TF.assign surprising behaviour
I wrote some tests to guide my understanding of `TF.assign` in relation to `run` and `runSession`. Similar to:
https://github.com/tensorflow/haskell/blob/master/tensorflow-ops/tests/BuildTest.hs#L83

```haskell
case_use_scalar_after_assigns_surprizingly_fails = do
  (f0, f1, f2) <- TF.runSession $ do
    (formula, reset0, reset1) <- do
      w <- TF.initializedVariable 0
      let formula = w                                                                                                                                                   
      r0 <- TF.assign w (TF.scalar (0.1::Float))
      r1 <- TF.assign w (TF.scalar (0.2::Float))
      return (formula, r0, r1)

    (TF.Scalar f0) <- TF.run formula
    TF.run_ reset0
    (TF.Scalar f1) <- TF.run formula
    TF.run_ reset1
    (TF.Scalar f2) <- TF.run formula
    return (f0, f1, f2)

  (0.0, 0.1, 0.2) @=? (f0, f1, f2)
```

This fails with f0, f1 and f2 being the final value of the assigned variable. If I change the test to:

```haskell
case_use_scalar_after_assigns_works_but_why = do
  (f0, f1, f2) <- TF.runSession $ do
    (formula, reset0, reset1) <- do
      w <- TF.initializedVariable 0
      let formula = TF.mul 1 w -- Matters
      r0 <- TF.assign w (TF.scalar (0.1::Float))
      r1 <- TF.assign w (TF.scalar (0.2::Float))
      return (formula, r0, r1)

    (TF.Scalar f0) <- TF.run formula
    TF.run_ reset0
    (TF.Scalar f1) <- TF.run formula
    TF.run_ reset1
    (TF.Scalar f2) <- TF.run formula
    return (f0, f1, f2)

  (0.0, 0.1, 0.2) @=? (f0, f1, f2)
```

by changing to `let formula = TF.mul 1 w` This works, but why doesn't the previous version?

I wrote these tests because I tried in other code do something like this:
```haskell
...
M.replicateM epochs $ do
    TF.run_ fitStep
    TF.run $ TF.value refW
```
to collect the evolution of the `w` terms but they ended up being equal (but verifiably different from the initial value), even with the `TF.value` conversion.
(The `fitStep` does `assign` to `refW`)

Is this expected behaviour? Am I missing something?

judah:
The main difference between these and the `Ref`-bases ops is the explicit
`readValue` op.  I'm not sure how this should interact with gradients
and save/restore, so I'm keeping it as a separate module for now.  Once we
figure out the details, we can merge it into `TensorFlow.Ops` and replace
all uses of the old `Ref`-based ops.  (Which will also fix #92.)

Also replaces our special case newtype `ResourceHandle` to
`Tensor Value ResourceHandle`, where `ResourceHandle` is the TF proto
corresponding to `DT_RESOURCE`.

### DIFF
--- a/tensorflow-ops/src/TensorFlow/Variable.hs
+++ b/tensorflow-ops/src/TensorFlow/Variable.hs
@@ -1,0 +1,109 @@
+-- | An implementation of ResourceHandle-based variables.
+--
+-- The main difference between this and 'Ref'-based variables is
+-- that reads are explicit, via the 'readValue' op.
+--
+-- TODO: given that distinction, figure out a good story around
+-- gradients and save/restore.  Then, merge this module into
+-- TensorFlow.Ops.
+{-# LANGUAGE RecursiveDo #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE OverloadedStrings #-}
+module TensorFlow.Variable
+    ( Variable
+    , variable
+    , variable'
+    , readValue
+    , initializedVariable
+    , initializedVariable'
+    , zeroInitializedVariable
+    , zeroInitializedVariable'
+    , assign
+    , assign'
+    , assignAdd
+    , assignAdd'
+    ) where
+
+import Data.Text.Encoding (encodeUtf8)
+import Lens.Family2 ((.~), (&))
+import TensorFlow.Core
+import TensorFlow.Build (opDef)
+import TensorFlow.BuildOp (buildInputs, pureOp, OpParams)
+import TensorFlow.Output (opInputs, unNodeName)
+import TensorFlow.Tensor (tensorNodeName)
+import TensorFlow.Types (tensorType)
+import qualified TensorFlow.GenOps.Core as CoreOps
+import TensorFlow.Ops (zeros)
+
+newtype Variable a = Variable (Tensor Value ResourceHandle)
+
+-- | Creates a new, uninitialized variable.
+variable :: (MonadBuild m, TensorType a) => Shape -> m (Variable a)
+variable = variable' id
+
+variable' :: forall m a . (MonadBuild m, TensorType a)
+                    => OpParams -> Shape -> m (Variable a)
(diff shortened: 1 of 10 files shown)


tensorflow-ops/src/TensorFlow/Variable.hs
```diff
@@ -0,0 +1,109 @@
...
+newtype Variable a = Variable (Tensor Value ResourceHandle)
+
+-- | Creates a new, uninitialized variable.
+variable :: (MonadBuild m, TensorType a) => Shape -> m (Variable a)
+variable = variable' id
+
+variable' :: forall m a . (MonadBuild m, TensorType a)
+                    => OpParams -> Shape -> m (Variable a)
+variable' params s = build $ do
+    -- Each variable needs a unique "shared_name".  Use MonadFix to
```
blackgnezdo: Thanks for this comment. The rec construct is otherwise obscure enough for people to stumble.

tensorflow-ops/src/TensorFlow/Variable.hs
```diff
@@ -0,0 +1,109 @@
...
+zeroInitializedVariable
+  :: (MonadBuild m, TensorType a, Num a) => Shape -> m (Variable a)
+zeroInitializedVariable = zeroInitializedVariable' id
+
+zeroInitializedVariable'
+  :: (MonadBuild m, TensorType a, Num a) => OpParams -> Shape -> m (Variable a)
+zeroInitializedVariable' params = initializedVariable' params . zeros
+
+-- | Gets the value stored in a variable.
+readValue :: TensorType a => Variable a -> Tensor Build a
```
fkm3: It is probably worth documenting that this is an odd duck. It is both stateful and subject to CSE.

PR #102, opened 2017-04-29
Implemented gradient of ReluGrad and Fill

jcberentsen:
Implementation of "ReluGradGrad", emulating the python implementation. Implementing gradient of "Fill" was also necessary to get the tests to run. Added two testcases, one for ReluGrad and one for ReluGradGrad.

I am unsure if implementations are correct when it comes to higher dimensionalities etc. Particularly the use of 'sum" in the "Fill" gradient as the python implementation uses math_ops.reduce_sum(grad).

### DIFF
--- a/tensorflow-ops/src/TensorFlow/Gradient.hs
+++ b/tensorflow-ops/src/TensorFlow/Gradient.hs
@@ -439,6 +439,7 @@ opGrad :: forall a . GradientCompatible a => Text -> GradientFunc a
 opGrad "Abs" _ [toT -> x] [dz] = [Just $ expr dz * signum x]
 opGrad "Neg" _ [_] [dz] = [Just $ negate $ expr dz]
 opGrad "Relu" _ [toT -> x] [dz] = [Just $ reluGrad dz x]
+opGrad "ReluGrad" _ [_, toT -> x ] [dz] = [Just $ reluGrad dz x, Just $ CoreOps.zerosLike x]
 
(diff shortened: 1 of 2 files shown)


blackgnezdo:
Jenkins, test this please.

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -667,6 +668,8 @@ opGrad "LabelClasses" _ _ _ = [Nothing, Nothing]
 opGrad "LabelWeights" _ _ _ = [Nothing]
 opGrad "Size" _ _ _ = [Nothing]
 opGrad "ZerosLike" _ _ _ = [Nothing]
+opGrad "Fill" _ _ [dz] = [Nothing, Just $ sum dz rx]
+  where rx = vector [0] :: Tensor Build Int32
```
fkm3: >I am unsure if implementations are correct when it comes to higher dimensionalities etc. Particularly the use of 'sum" in the "Fill" gradient as the python implementation uses math_ops.reduce_sum(grad).

I think you want something like this

```haskell
sum dz (range 0 (rank dz) 1)
```

jcberentsen: Thanks, I see there is a function `rangeOfRank` in `Gradient.hs`, I'll use that.

blackgnezdo:
Jenkins, test this please.

fkm3:
Thanks for the contribution!

PR #107, opened 2017-05-08
Increase the number of iterations for MatrixTest.

judah:
The number of iterations was reduced from 1000 to 300 during review, but that
turned out to be too low and the test now fails about 20% of the time.
After changing it back to 1000, the test succeeded at 50 out of 50 runs.


judah:
Jenkins, test this please.

judah:
Jenkins, test this please

blackgnezdo:
Jenkins, test this please

PR #108, opened 2017-05-09
Work around #92 by always copying TensorData when fetching.

judah:
It would be better to avoid the copy when it's not necessary, but
that will require more involved changes to the internal API.  (For example,
Fetchable might need to allow IO or ST actions.)

### DIFF
--- a/tensorflow-ops/tests/OpsTest.hs
+++ b/tensorflow-ops/tests/OpsTest.hs
@@ -94,10 +94,22 @@ testScalarFeedCse = testCase "testScalarFeedCse" $ TF.runSession $ do
                 $ p1 `TF.add` p2
     liftIO $ result @=? TF.Scalar 5
 
+-- | See https://github.com/tensorflow/haskell/issues/92.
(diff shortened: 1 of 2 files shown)


fkm3:
Jenkins, test this please.

fkm3:
How would Fetchable avoid the copy with IO or ST? Don't you need to know whether the TensorData pointer is alias-able within the tensorflow implementation to decide whether to copy?

judah:
If the output type requires a copy anyway (e.g. `[Int64]` or `Data.Unboxed.Vector Float`), then this change causes the data to be copied twice: once to get a `Storable.Vector`, and once to get the final type.

Instead,c`TensorData` itself would have to wrap an `S.IOVector` instead of an `S.Vector`, and the `Fetchable` action could run in `IO` to extract the final type (with special logic for each type's instance to decide how much copying to do).

fkm3:
Ah, makes sense.

I think you need to rebase/merge HEAD to fix the tests.

PR #110, opened 2017-05-10
Fix .cabal files so 'stack check' passes.

judah:
- Add LICENSE files for all packages.
- Add descriptions for packages that were missing one.
- Work around google/proto-lens#69 by symlinking third_party into
  tensorflow-proto.

### DIFF
--- a/tensorflow-proto/tensorflow-proto.cabal
+++ b/tensorflow-proto/tensorflow-proto.cabal
@@ -4,16 +4,17 @@ synopsis:            TensorFlow protocol buffers.
 description:         Please see README.md
 homepage:            https://github.com/tensorflow/haskell#readme
 license:             Apache
+license-file:        LICENSE
 author:              TensorFlow authors
 maintainer:          tensorflow-haskell@googlegroups.com
 copyright:           Google Inc.
 category:            Machine Learning
 build-type:          Custom
 cabal-version:       >=1.24
-extra-source-files:  ../third_party/tensorflow/tensorflow/core/framework/*.proto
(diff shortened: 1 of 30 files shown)


tensorflow-proto/tensorflow-proto.cabal
```diff
@@ -4,16 +4,17 @@ synopsis:            TensorFlow protocol buffers.
...
 maintainer:          tensorflow-haskell@googlegroups.com
 copyright:           Google Inc.
 category:            Machine Learning
 build-type:          Custom
 cabal-version:       >=1.24
-extra-source-files:  ../third_party/tensorflow/tensorflow/core/framework/*.proto
-                   , ../third_party/tensorflow/tensorflow/core/protobuf/config.proto
-                   , ../third_party/tensorflow/tensorflow/core/protobuf/debug.proto
-                   , ../third_party/tensorflow/tensorflow/core/util/event.proto
+extra-source-files:  third_party/tensorflow/tensorflow/core/framework/*.proto
```
blackgnezdo: How will this work for stand-alone cabal builds? Are you planning to include these files into the package?

judah: Yes, for this first release I was planning to include the .proto files in the tarball we upload to Hackage.

Longer-term I'd like to upload the generated `*.hs` files themselves, to avoid the dependency on `protoc`.  But that will require more changes to `Data.ProtoLens.Setup`.

judah: (To clarify, this new setup causes the linked `.proto` files to be included in the tarball automatically.)

PR #113, opened 2017-05-10
Fix 'sdist' for tensorflow-core-ops.

judah:
Use the same trick as for `proto-lens-protoc`: hack the package description
during the `sdist` step to include the autogen directory in hs-source-dirs.


judah:
For reference, this is the corresponding code in proto-lens-protoc:
https://github.com/google/proto-lens/blob/master/proto-lens-protoc/src/Data/ProtoLens/Setup.hs#L145

PR #116, opened 2017-05-12
Support Variable in TensorFlow.Gradient and use in mnist example

fkm3:
Maybe a more general "TensorLike" type class would make more sense than `GradientTarget`? Let me know what you think.

For the remaining `Variable` blockers: I think it would make sense to only support `Variable` in our `save` and `restore` wrappers.


tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -171,10 +181,9 @@ gradients y xs = build $ do
     gradientMap <- graphGrads gr initPending
     -- Lookup the gradients for each x.
     forM xs $ \x ->
-        let xName = tensorNodeName x
-        in maybe (render $ zerosLike x) return $ do
+        let (Output i xName) = targetOutput x
```
judah: Remove redundant parentheses around `Output`?

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -687,9 +696,12 @@ opGrad "Fill" _ _ [dz] = [Nothing, Just $ sum dz rx]
   where
     rx = rangeOfRank dz
 
+opGrad "ReadVariableOp" _ _ [dz] = [Just $ expr dz]
```
judah: Please add a comment clarifying how this works, since it took me a bit to wrap my head around.

My understanding is: because the only input of `ReadVariableOp` is the variable itself, the gradient with respect to the read gets back propagated to the gradient w.r.t the variable.  And the variable node itself has no inputs, so no more backdrop takes place.

tensorflow-ops/src/TensorFlow/Variable.hs
```diff
@@ -30,13 +30,18 @@ import TensorFlow.Core
...
 import qualified TensorFlow.GenOps.Core as CoreOps
-import TensorFlow.Ops (zeros)
+import TensorFlow.Ops (zeros, fill, shape)
+import TensorFlow.Gradient (GradientTarget(..))
 
 newtype Variable a = Variable (Tensor Value ResourceHandle)
 
+instance GradientTarget Variable where
+    targetOutput (Variable v) = renderedOutput v
+    targetZeros (Variable v) = fill (shape v) 0
```
judah: I *think* this won't do what's intended.  A Variable is a multidimensional array of `DT_RESOURCE` type whose dimensions happen to contain only one element, so as far as I can tell this will always return a constant size.

See for example the implementation of ReadHandle:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/resource_variable_ops.cc#L44
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/resource_mgr.cc#L246

I think we want `zerosLike (readValue v)` here instead?  Which might suggest that the class method we want is closer to (`t a -> Tensor Value a`)?

(Even better if you can add a test that confirms whether my understand is correct.)

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -108,6 +109,15 @@ type GradientCompatible a =
     -- TODO(fmayle): MaxPoolGrad doesn't support Double for some reason.
     (Num a, OneOf '[ Float, Complex Float, Complex Double ] a)
 
+-- | Types that gradients can be taken w.r.t. (namely 'Tensor' and 'Variable').
+class GradientTarget t where
```
judah: To the question of generalizing the name, TensorFlow has `tf.convert_to_tensor` which seems like a similar effect:
https://www.tensorflow.org/api_docs/python/tf/convert_to_tensor
It feels a little odd to put something as specific as `targetZeros` in a general class like that, though.  Up to you; see my comment below in `TensorFlow.Variable`.

judah:
Also, +1 to making save/restore only work with `Variable`s.

tensorflow-ops/src/TensorFlow/Gradient.hs
fkm3: I agree `targetZeros` doesn't make much sense in a general purpose class since `shape (targetZeros x) == shape x` isn't always true. If there is another use case for getting the Output of a tensor/variable, then maybe moving `targetOutput` to a separate class would be useful. Sticking with `GradientTarget` for now.

tensorflow-ops/src/TensorFlow/Gradient.hs
fkm3: Done.

tensorflow-ops/src/TensorFlow/Gradient.hs
fkm3: Right. Added a comment which hopefully makes it clearer.

tensorflow-ops/src/TensorFlow/Variable.hs
fkm3: You're right. Added a test case + converted all of the existing gradient test cases to use `Variable`.

tensorflow-ops/src/TensorFlow/Gradient.hs
fkm3: One use case for something like `targetOutput` would be `colocateWith`.

fkm3: `targetOutput` is really just `renderedOutput` (except lifted inside `ResourceHandle` wrapper for the non-`Tensor` cases), so I'm thinking of adding a class for `renderedOutput`. One hesitation is that it overlaps with `Rendered` which would likely be confusing to users, but I think we can replace `Rendered` completely with the `renderedOutput` class.

For `targetZeros`, I like your `tf.convert_to_tensor` suggestion after thinking about it more. I'll try that and do `zerosLike (convertToTensor x)`.

judah: SGTM

tensorflow-ops/src/TensorFlow/Variable.hs
fkm3: I dropped the new test case since it seems too trivial now (it just does readValue).

fkm3:
PTAL. I've incorporated the changes from the other two PRs.

judah:
Jenkins, test this please.

PR #121, opened 2017-05-16
Add ToTensor class

fkm3:


tensorflow/src/TensorFlow/Tensor.hs
```diff
@@ -189,3 +190,11 @@ instance TensorKind Ref where
 
 instance TensorKind Build where
     toBuild = id
+
+
+-- | Types which can be converted to `Tensor`.
+class ToTensor t where
```
judah: Would it be helpful to make `ToTensor` a superclass of `Rendered`. to reduce the signature of consumer functions like `gradients`?  Up to you.

fkm3: I was thinking this might include non-`Rendered` stuff, like `TensorType a => [a]`. I actually tried to go a bit farther and make it work for scalars as well, but it seems to break type inference for numeric literals which kind of defeats the purpose.

Issue #118, opened 2017-05-12
Remove the dependency on snappy


judah:
Our bindings currently link against the `snappy` C library due to the dependency on `snappy-framing`.  As far as I can tell, the only reason we use that library is for `Data.Digest.CRC32C` which is pure Haskell code and doesn't actually use that C dependency.

The `snappy` and `snappy-framing` packages are also currently not in Stackage, which prevents `tensorflow-records` and `tensorflow-records-conduit` from being there as well.

As suggested in the code, one option is to split out the crc32c functionality into its own package.

PR #122, opened 2017-05-16
Add extra-lib-dirs back into .cabal for Mac builds.

judah:
Also bump the version to 0.1.0.2.

Originally we had `extra-lib-dirs: /usr/local/lib` in `stack.yaml`.
I removed it because it wasn't necessary on my Mac.  However,
it turns out that it is necessary for machines with the default installation
of XCode, which *doesn't* search that path by default.

(On my machine, it wasn't necessary because I had run `xcode-select --install`
which adds that path permanently to your search path.)

I'm adding the setting back to `tensorflow.cabal` as well as `stack.yaml` so
that the Hackage release also contains this fix.  Changing `stack.yaml` is
still necessary in order to fix linkage in the `snappy` package (which
`tensorflow-records` depends on).  Hopefully that will go away once we remove
the dependency (#118).

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -16,3 +16,11 @@ packages:
 extra-deps:
 - snappy-framing-0.1.1
 - snappy-0.2.0.2
+
+# For Mac OS X, whose linker doesn't use this path by default
+# unless you run `xcode-select --install`.
+# TODO: remove this once we stop depending on `snappy`.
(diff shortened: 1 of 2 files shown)


judah:
For more context, see Homebrew/brew#556.  Long-term I'll look into submitting a patch for one of GHC/Cabal/Stack.

PR #123, opened 2017-05-16
Bump version to 0.2.0.0, and start a changelog.

judah:
The HEAD contains #119 which is an API change requiring (as far as I can tell)
a major version bump.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,7 +1,9 @@
 # ChangeLog
 
 ## Upcoming (v0.2.0.0)
-- Expand the Rendered class to support ResourceHandle wrappers like Variables.
+- Expand the `Rendered` class and add a `ToTensor` class to let more functions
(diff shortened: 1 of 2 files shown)


judah:
@fkm3 , I assume when we release `v0.2.0.0` it will contain all the changes you have queued up around `Variables` and `ToTensor`.  (This PR is mainly about bookkeeping for the future.)

Separately, I've released `v0.1.0.2` on Hackage to fix the previous OS X build problem.  That release contains all of HEAD except for your change to the `Rendered` class, since the latter will require a major version bump.  The release version is tagged `v0.1.0.2` and has branch name `release-0.1.0.2`.

fkm3:
LGTM, but I think you need to bump the version in the dependent packages to fix the build.

judah:
Reverted the `.cabal` change; we can do that for the actual release, since the exact set of changes included in the release will determine which packages need to be bumped for PVP concerns.

Also updated the changelog to cover more of your recent PRs.

PR #124, opened 2017-05-18
Add initializedValue function for Variable

fkm3:


tensorflow-ops/src/TensorFlow/Variable.hs
```diff
@@ -30,15 +31,16 @@ import TensorFlow.Core
...
 import TensorFlow.Output (opInputs, unNodeName)
-import TensorFlow.Tensor (Rendered(..), ToTensor(..), tensorNodeName)
+import TensorFlow.Tensor (Rendered(..), ToTensor(..), renderValue, tensorNodeName)
 import TensorFlow.Types (tensorType)
 import qualified TensorFlow.GenOps.Core as CoreOps
 import TensorFlow.Ops (zeros)
 
-newtype Variable a = Variable (Tensor Value ResourceHandle)
+data Variable a =
+    Variable (Tensor Value ResourceHandle) (Maybe (Tensor Value a))
```
judah: Add a comment saying what each field is, and/or make this a record type.

fkm3: Done.

PR #126, opened 2017-05-28
Modifying tensorflow-mnist-input-data's setup to download MNIST through http proxy if necessary

helq:
Changing way to download MNIST dataset from Network.HTTP.simpleHTTP to Network.Browser.browse which allows to capture the environment's proxy.


helq:
I signed it!

PR #132, opened 2017-06-06
Fill out the modules in tensorflow-proto.

judah:
That package now includes everything in the `tensorflow/core:protos_all` target:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/BUILD#L131

I also made all of the modules exposed for simplicity.  (As a particular
example: `MetaGraph`, which was previously in `other-modules`, is useful for
constructing some TensorFlow serving RPCs.)

### DIFF
--- a/tensorflow-proto/tensorflow-proto.cabal
+++ b/tensorflow-proto/tensorflow-proto.cabal
@@ -11,32 +11,54 @@ copyright:           Google Inc.
 category:            Machine Learning
 build-type:          Custom
 cabal-version:       >=1.24
-extra-source-files:  third_party/tensorflow/tensorflow/core/framework/*.proto
-                   , third_party/tensorflow/tensorflow/core/protobuf/config.proto
-                   , third_party/tensorflow/tensorflow/core/protobuf/debug.proto
-                   , third_party/tensorflow/tensorflow/core/util/event.proto
+extra-source-files:  third_party/tensorflow/tensorflow/core/example/*.proto
+                   , third_party/tensorflow/tensorflow/core/framework/*.proto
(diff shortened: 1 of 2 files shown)


judah:
Note: this is blocked on a new release of `proto-lens` that includes google/proto-lens#81 (and, possibly, google/proto-lens#83).

tensorflow-proto/tensorflow-proto.cabal
```diff
@@ -11,32 +11,54 @@ copyright:           Google Inc.
...
+                     , Proto.Tensorflow.Core.Protobuf.SavedModel
+                     , Proto.Tensorflow.Core.Protobuf.Saver
+                     , Proto.Tensorflow.Core.Protobuf.TensorflowServer
+                     , Proto.Tensorflow.Core.Util.Event
+                     , Proto.Tensorflow.Core.Util.MemmappedFileSystem
+                     , Proto.Tensorflow.Core.Util.SavedTensorSlice
+                     , Proto.Tensorflow.Core.Util.TestLog
   build-depends:  proto-lens == 0.2.*
                 , proto-lens-protoc == 0.2.*
+                , proto-lens-protobuf-types == 0.2.*
```
blackgnezdo: Presumably this is the new package which will be released with the next proto-lens?

judah: Yep; they're released now.  Updated `stack.yaml` to use the new versions.

Sometime next week there will be a new `lts-8.*` release with these new `proto-lens-*` versions, which will let us simplify the stack file further (other than `proto-lens-protobuf-types`, which might need to wait for the next LTS version`).

PR #142, opened 2017-07-17
Change from OS X to macOS

alanyee:
Couple of changes from OS X to macOS


alanyee:
@judah I have done so.

alanyee:
@judah does everything look good?

PR #154, opened 2017-09-29
Create monad transformer version of Session (closes #153)

### LINKED ISSUE #153: Monad transformer version of Session
It would be useful to have a `SessionT` as well, such that we can run the TensorFlow operations in any monad with a `MonadIO` instance instead of just `IO` itself.

jbransen:
`Session` was already a monad transformer stack, which is now also exported. All other signatures have become a bit more generic but are all backwards compatible.

### DIFF
--- a/tensorflow/src/TensorFlow/Internal/FFI.hs
+++ b/tensorflow/src/TensorFlow/Internal/FFI.hs
@@ -33,8 +33,9 @@ module TensorFlow.Internal.FFI
 
 import Control.Concurrent.Async (Async, async, cancel, waitCatch)
 import Control.Concurrent.MVar (MVar, modifyMVarMasked_, newMVar, takeMVar)
-import Control.Exception (Exception, throwIO, bracket, finally, mask_)
 import Control.Monad (when)
+import Control.Monad.Catch (MonadMask, Exception, throwM, bracket, finally, mask_)
+import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Bits (Bits, toIntegralSized)
 import Data.Int (Int64)
 import Data.Maybe (fromMaybe)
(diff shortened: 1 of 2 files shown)


jbransen:
I signed it!

fkm3:
Jenkins, test this please.

fkm3:
@judah does this seem reasonable to you?

judah:
Yeah, makes sense to me.

Looks like the CLA robot is stalled though?

fkm3:
Yeah, from my side it looks like the CLA hasn't been signed yet. @jbransen, can you give it another try?

jbransen:
My employer has signed, I've double checked that the right Google group is entered and that the e-mailaddress I used for the commit is indeed in that group. My employer and me can't really check anything else, as the CLA website only says "We don't have any individual agreements on file for you." (which is correct, since it is not an individual agreement), and we get no confirmation after te form has been signed. Is there anything else I can check?

fkm3:
Ah, was the corp CLA signed recently? The docs say that, unlike individual CLAs, they may take a few days to process.

jbransen:
Yes, a few minutes before the pull request. I don't see anywhere that it could take a few days, but let's see then.

tensorflow/src/TensorFlow/Internal/FFI.hs
```diff
@@ -75,25 +76,26 @@ data TensorData = TensorData
...
                 checkStatus (Raw.deleteSession s)
-    bracket Raw.newSessionOptions Raw.deleteSessionOptions $ \options -> do
-        optionSetter options
+    bracket (liftIO $ Raw.newSessionOptions) (liftIO . Raw.deleteSessionOptions) $ \options -> do
+        liftIO $ optionSetter options
         bracket
-            (checkStatus (Raw.newSession options))
-            cleanup
+            (liftIO $ checkStatus (Raw.newSession options))
+            (liftIO . cleanup)
```
blackgnezdo: 5 liftIO in 4 lines feels like an overkill. Can this code be refactored a bit to reduce verbiage? It feels almost like you can pull this outside if it were possible to "unliftIO" the line below. Not that I'm advocating adding a dependency on 
https://hackage.haskell.org/package/unliftio ...

fkm3: Here is one idea:
```haskell
    let bracketIO x y = bracket (liftIO x) (liftIO . y)
    bracketIO Raw.newSessionOptions Raw.deleteSessionOptions $ \options -> do
        bracketIO
            (optionSetter options >> checkStatus (Raw.newSession options))
            cleanup
            (action (asyncCollector drain))
```

fkm3:
The doc I saw may only be available internally. I'll ping someone internally if the CLA doesn't go through in a day or two.

tensorflow/src/TensorFlow/Internal/FFI.hs
blackgnezdo: Yeah, this is much nicer. Let's make it so :)

jbransen:
I signed it!

jbransen:
So, I did the suggestion code change (indeed a bit nicer), and also just got the confirmation that the CLA is now accepted.

fkm3:
Jenkins, test this please.

fkm3:
It tells me the CLA isn't signed for your github username, but that it is signed for your email. I think there is just some latency in the system, so I'm going to ignore it and submit. Thanks for the contribution!

PR #155, opened 2017-10-11
Gradient of Conv2DBackpropInput

jcberentsen:


blackgnezdo:
Jenkins test this please.

tensorflow-ops/tests/GradientTest.hs
```diff
@@ -32,7 +32,7 @@ import Control.Monad(forM_, replicateM, zipWithM)
 import Control.Monad.IO.Class (liftIO)
 
 import qualified TensorFlow.Core as TF
-import qualified TensorFlow.GenOps.Core as TF (max, tile, maximum)
+import qualified TensorFlow.GenOps.Core as TF (max, tile, maximum, conv2DBackpropInput')
```
blackgnezdo: Could you keep the imports sorted in ()? I know they weren't before...

tensorflow-ops/tests/GradientTest.hs
```diff
@@ -388,6 +389,28 @@ transAttrs :: (TF.Attribute a,
...
+        let conv_input_shape = TF.vector [1, 2, 2, 1 :: Int32] -- [batch, h, w, in_channels]
+        let conv_out_shape = TF.vector [1, 1, 1, 1 :: Int32]  -- [batch, h, w, out_channels]
+        x <- TF.render $ TF.fill conv_out_shape (TF.scalar (1::Float))
+
+        let filterShape = TF.vector [2, 2, 1, 1 :: Int32] -- [fh, fw, inc, out]
+        filter <- TF.render $ TF.fill filterShape (TF.scalar (1::Float))
+        let y = TF.conv2DBackpropInput'
+                ( (TF.opAttr "strides" .~ [1::Int64, 1, 1, 1])
+                . (TF.opAttr "padding" .~ (BS.pack "VALID"))
+                . (TF.opAttr "data_format" .~ (BS.pack "NHWC"))
```
blackgnezdo: This is a fairly inconvenient API. Would you be open to putting some types around TF.conv2DBackpropInput' to make it more friendly? A couple of sum types could go a long way in improving usability. Maybe something in TensorFlow.Core? Unless @fkm3 has a better idea for placing it.

fkm3: If you wanted to add a wrapper, I'd suggest one of the tensorflow-ops modules. In this case it probably isn't worth the overhead since no one should want to call TF.conv2DBackpropInput directly (@jcberentsen must be calculating second derivatives or something, correct me if I'm worng). Nicer wrappers for conv2d/etc would definitely be worth it though, I have half baked idea for that here https://github.com/fkm3/tensorflow-haskell/blob/dqn/tensorflow-mnist/app/Main.hs#L70 (implementation here https://github.com/fkm3/tensorflow-haskell/blob/dqn/tensorflow-ops/src/TensorFlow/Layer.hs#L106).

fkm3:
The changes LGTM, thanks!

tensorflow-ops/tests/GradientTest.hs
blackgnezdo: Resolving this then.

jcberentsen: `TF.conv2DBackpropInput` was useful for a CNN based AutoEncoder, where deconv is `conv2d_transposed` which is implemented in terms of conv2DBackpropInput. The ideas outlined by @fkm3 looks better than my own local wrappers for conv2D (and conv2d_transpose)

I am interested in the `Layer.hs` approach. I think it also needs some way to control variable reuse (for GAN implementations) I have some sketches for some rudimentary variable reuse in a WGAN implentation but I am not happy with it. Basically passing variables to reuse with `Maybe [TF.Variable a]` but this needs to scale to reusing variables for multiple layers.

blackgnezdo:
Jenkins test this please.

tensorflow-ops/tests/GradientTest.hs
fkm3: You can get variable reuse like this, but it is a bit awkward
```haskell
    (layer1params, _, layer1) <- Layer.build [...] (Layer.dense 1024 TF.relu)
    (remainingParams, _, nn) <- Layer.build [...] $
        Layer.liftPure layer1
        >>> ...
        >>> Layer.liftPure layer1
        >>> ...
``

PR #159, opened 2017-10-16
Added imageSummary wrapper

jcberentsen:


tensorflow-logging/src/TensorFlow/Logging.hs
```diff
@@ -58,7 +61,8 @@ import Data.ByteString (ByteString)
 import Data.Conduit ((=$=))
 import Data.Conduit.TQueue (sourceTBMQueue)
 import Data.Default (def)
-import Data.Int (Int64)
+import Data.Int
```
blackgnezdo: Could you spell out the imports?

tensorflow-logging/src/TensorFlow/Logging.hs
```diff
@@ -151,16 +155,27 @@ doubleWallTime = asDouble <$> getCurrentTime
 -- | Adds a 'CoreOps.histogramSummary' node. The tag argument is intentionally
 -- limited to a single value for simplicity.
 histogramSummary ::
-    (MonadBuild m, TensorType t, t /= ByteString, t /= Bool)
-     -- OneOf '[Int16, Int32, Int64, Int8, Word16, Word8, Double, Float] t)
+    (MonadBuild m, TensorType t, OneOf '[Int16, Int32, Int64, Int8, Word16, Word8, Double, Float] t)
```
blackgnezdo: Why is this one preferable to the shorter above? (The comment was redundant, not sure why we keep it)

tensorflow-logging/src/TensorFlow/Logging.hs
```diff
@@ -151,16 +155,27 @@ doubleWallTime = asDouble <$> getCurrentTime
...
+    let stepString = CoreOps.asString step
+        tag = scalarize $ CoreOps.add (scalar label) stepString
+    addSummary . CoreOps.imageSummary tag
+
 -- | Adds a 'CoreOps.scalarSummary' node.
 scalarSummary ::
-    (TensorType t, t /= ByteString, t /= Bool, MonadBuild m)
-    -- (TensorType t,
-    --  OneOf '[Int16, Int32, Int64, Int8, Word16, Word8, Double, Float] t)
+    (TensorType t, OneOf '[Int16, Int32, Int64, Int8, Word16, Word8, Double, Float] t, MonadBuild m)
```
blackgnezdo: Same question as about L158.

tensorflow-logging/src/TensorFlow/Logging.hs
jcberentsen: Sure

tensorflow-logging/src/TensorFlow/Logging.hs
jcberentsen: I read the comment to indicate something wanting to be expressed, so I turned it into code instead. Agree it is shorter the way it was, but less explicit. So should I revert it and delete the comment?

tensorflow-logging/src/TensorFlow/Logging.hs
```diff
@@ -151,16 +154,27 @@ doubleWallTime = asDouble <$> getCurrentTime
...
+    (MonadBuild m, OneOf '[Int16, Int32, Int64, Int8, Word16, Word8, Double, Float] t)
     => ByteString -> Tensor v t -> m ()
 histogramSummary tag = addSummary . CoreOps.histogramSummary (scalar tag)
 
+-- | Adds a 'CoreOps.imageSummary' node. The tag used in the summary is
+-- | computed from label <> step.
+imageSummary ::
+    (OneOf '[Word8, Word16, Float] t, MonadBuild m)
+    => ByteString -- label
+    -> Tensor v1 Int64 -- step
```
fkm3: Does it make sense to include step in the tag automatically? That seems like something the caller can do themselves.

tensorflow-logging/src/TensorFlow/Logging.hs
fkm3: I think the intent is to allow any numeric tensortype, so the old way is somewhat more future proof assuming no more non-numeric types are added. I don't have a strong opinion on which way is better.

tensorflow-logging/src/TensorFlow/Logging.hs
fkm3: Does it make sense to include step in the tag automatically? That seems like something the caller can do themselves.

jcberentsen: Agree, the user can do that. I'll remove it. Maybe add a hint on how to construct a numbered label, or a helper function for that somewhere?

blackgnezdo:
Jenkins please test this.

jcberentsen:
Any chance to get this merged?

fkm3:
Merged. blackgnezdo@ approved the PR, which I think gave you the ability to merge it. Sorry if there was confusion.

PR #168, opened 2017-11-12
Use stack's nix support

nmattia:
This makes use of `stack --nix` instead of providing `tools/userchroot`. It provides a `shell.nix` that can be used but now the users can build with `stack --nix build`. For me `stack --nix test` works out of the box. The nix support is disabled by default in the `stack.yaml`, hence you _do_ need to pass `--nix` to your stack commands.

This fixes a few issues I ran into while trying to build:
 * The nixpkgs version is pinned, with a script `nix/update` that will update the commit to the header of `release-17.09`.
 * The tensorflow library used is patched to ensure the libraries (libc, libstdc++) are present on the system.

AFAICT this removes the need for `buildFHSUserEnv`.

Now I need to figure out what that Google CLA is about...


blackgnezdo:
Thanks for the patch. Please sign CLA.

@grwlf who added nix support in the first place should probably take a look as I don't understand nor use Nix.

nmattia:
I signed it!

blackgnezdo:
Jenkins test this please.

traleig1:
This works fantastically, thanks

jyp:
What is blocking this to be merged? (Apart from the obvious conflict?) @grwlf ?

fkm3:
I think we just want someone else nix literate to take a look. If you'd like to look over the code and give a thumbs up that should be good enough :)

jyp:
This PR is a definite improvement over what's currently in master. So I recommend merging it.
 It needs to be rebased though. Mainly the version of tensorflow-c should be updated so that it would match the proto. Perhaps @nmattia can confirm.

fkm3:
Thanks! @nmattia do you mind rebasing? (I think you just need to update the version in the tarball URL + its SHA)

nmattia:
Yep, will do. I should have a few free cycles tomorrow.

nmattia:
I've:
 * Updated to tensorflow 1.8.0
 * Added stack to the nix-shell (so that you don't even need stack installed)
 * Updated nixpkgs
 * Set GHC to 8.2.2 (the latest in nixpkgs is > 8.2.2 which doesn't work with the LTS)
 * Rebased

@jyp can you let me know if this still works for you? I don't have a global stack install, and adding stack to the nix-shell may cause issues there.

jyp:
@nmattia Thanks a lot; this is a very useful PR. (I don't have a global stack install either.)

PR #161, opened 2017-10-19
Update to tensorflow 1.3

fkm3:
* Tested on linux without Docker.
* Couldn't get nix build to work, so I just updated the URL and hash.
* Did not test macos build.

The mnist change was necessary because the argmax output type is now polmorphic.

PR #178, opened 2018-03-07
Update to tensorflow 1.6

jbransen:


blackgnezdo:
Thanks for PR, there are other places that need adjustment.

I recommend looking at #161 for a template.

jbransen:
You're right, I'll look into it.

jbransen:
Sorry, it's not going to work. For my project I only need `tensorflow-proto` and `tensorflow`, for which this works fine, but due to some changes the opsgen breaks, in ways that I can't easily seem to fix. And in the mean time there is already 1.7 RC1, so it makes sense to wait for the 1.7 release before updating the Haskell bindings now.

fkm3:
I think I've worked out the kinks for 1.7. The changes are at https://github.com/fkm3/tensorflow-haskell/tree/tensorflow_1_7

Will sent for review after https://github.com/tensorflow/haskell/pull/184 is in.

PR #184, opened 2018-04-15
Fix initialized variables for tensorflow 1.7

fkm3:
This is needed to support tensorflow 1.7. The trick of initializing a
variable with `Shape []` and then overriding the shape by assigning an
initial value no longer works. It seems that we need to explicitly flip
the `unknown_rank` bit in the shape proto.

I thought about switching opgen to use `Maybe Shape` when an op requires
a shape attribute, but that will cause a lot of api churn, so I chose to
hold off for now and just do a spot fix to unblock 1.7.

The new test assert already passes in the current version, but fails in 1.7.

### DIFF
--- a/tensorflow/src/TensorFlow/Types.hs
+++ b/tensorflow/src/TensorFlow/Types.hs
@@ -69,12 +69,13 @@ import Data.Default (def)
 import Data.Int (Int8, Int16, Int32, Int64)
 import Data.Maybe (fromMaybe)
 import Data.Monoid ((<>))
+import Data.ProtoLens.TextFormat (showMessageShort)
 import Data.Proxy (Proxy(..))
 import Data.String (IsString)
 import Data.Word (Word8, Word16, Word64)
 import Foreign.Storable (Storable)
 import GHC.Exts (Constraint, IsList(..))
-import Lens.Family2 (Lens', view, (&), (.~))
+import Lens.Family2 (Lens', view, (&), (.~), (^..))
 import Lens.Family2.Unchecked (iso)
 import Text.Printf (printf)
 import qualified Data.Attoparsec.ByteString as Atto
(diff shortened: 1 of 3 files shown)


tensorflow/src/TensorFlow/Types.hs
```diff
@@ -363,8 +368,23 @@ instance IsList Shape where
 protoShape :: Lens' TensorShapeProto Shape
 protoShape = iso protoToShape shapeToProto
   where
-    protoToShape = Shape . fmap (view size) . view dim
-    shapeToProto (Shape ds) = (def :: TensorShapeProto) & dim .~ fmap (\d -> def & size .~ d) ds
+    protoToShape p = fromMaybe (error msg) (view protoMaybeShape p)
+      where msg = "Can't convert TensorShapeProto with unknown rank to Shape"
```
blackgnezdo: Anything out of p that you can including into msg to help with diagnostics? ++ show p?

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -363,8 +368,23 @@ instance IsList Shape where
...
+    protoToShape :: TensorShapeProto -> Maybe Shape
+    protoToShape p =
+        if view unknownRank p
+            then Nothing
+            else Just (Shape (view size <$> view dim p))
+    shapeToProto :: Maybe Shape -> TensorShapeProto
+    shapeToProto Nothing =
+        def & unknownRank .~ True
+    shapeToProto (Just (Shape ds)) =
+        def & dim .~ fmap (\d -> def & size .~ d) ds
```
blackgnezdo: It feels like traverse could be used here?

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -363,8 +368,23 @@ instance IsList Shape where
...
+    shapeToProto s' = def & protoMaybeShape .~ Just s'
+
+protoMaybeShape :: Lens' TensorShapeProto (Maybe Shape)
+protoMaybeShape = iso protoToShape shapeToProto
+  where
+    protoToShape :: TensorShapeProto -> Maybe Shape
+    protoToShape p =
+        if view unknownRank p
+            then Nothing
+            else Just (Shape (view size <$> view dim p))
```
blackgnezdo: Is this p ^.. dim.size?

tensorflow/src/TensorFlow/Types.hs
blackgnezdo: I think what you have is good.

PR #185, opened 2018-04-17
Update to tensorflow 1.7

fkm3:
All of the non-`s/1.3/1.7/` changes are because
* There are new tensorflow datatypes
* Some ops have looser types (e.g. `fill` now accepts both int64 and int32)
* There are more ops of type "func"

### DIFF
--- a/tensorflow/src/TensorFlow/Types.hs
+++ b/tensorflow/src/TensorFlow/Types.hs
@@ -41,7 +41,7 @@ module TensorFlow.Types
     , Attribute(..)
     , DataType(..)
     , ResourceHandle
-    , Variant(..)
+    , Variant
     -- * Lists
     , ListOf(..)
     , List
@@ -129,7 +129,7 @@ type ResourceHandle = ResourceHandleProto
 -- | Dynamic type.
 -- TensorFlow variants aren't supported yet. This type acts a placeholder to
 -- simplify op generation.
-data Variant = Variant
+data Variant
 
 -- | The class of scalar types supported by tensorflow.
 class TensorType a where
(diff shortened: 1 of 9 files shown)


tools/userchroot.nix
```diff
@@ -13,8 +13,8 @@ let
     name = "tensorflow-c";
 
     src = fetchurl {
-      url = "https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-linux-x86_64-1.3.0.tar.gz";
-      sha256 = "1d4bda5316063b70cf50a668d774b2067ef2a8ab163ff2eb29592bf3c24e2183";
+      url = "https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-linux-x86_64-1.7.0.tar.gz";
+      sha256 = "621642b1fddd3831e048817d2220d9d7cf8ba359ac81c83a808bcdd9a982ee90";
```
blackgnezdo: I don't suppose you actually tested nix?

tensorflow/src/TensorFlow/Types.hs
```diff
@@ -124,6 +126,11 @@ import qualified TensorFlow.Internal.FFI as FFI
 
 type ResourceHandle = ResourceHandleProto
 
+-- | Dynamic type.
+-- TensorFlow variants aren't supported yet. This type acts a placeholder to
+-- simplify op generation.
+data Variant = Variant
```
blackgnezdo: Is void type at all plausible here? I.e. 
data Variant

tools/userchroot.nix
fkm3: No, I get an error when I try the build instructions. I figured it was worth updating the url and hash though.

tensorflow/src/TensorFlow/Types.hs
fkm3: Yeah. I've removed the data constructor.

fkm3:
Thanks for the reviews!

PR #189, opened 2018-05-15
Use newer version of stack in CI

fkm3:
Required by #187.

The version we were using is old enough that it doesn't work with the
latest stackage LTS. haskellstack.org says

    There is also a Ubuntu package for Ubuntu 16.10 and up, but the
    distribution's Stack version lags behind, ...

So, instead of asking them to update it, it's probably better to
download the tar of the version we want.

Somehow updating stack surfaced a new pedantic warning in GradientTest,
so I've fixed that as well (and cleaned up some code using snake_case).

### DIFF
--- a/ci_build/Dockerfile
+++ b/ci_build/Dockerfile
@@ -12,6 +12,7 @@ ADD . /tfhs
 WORKDIR /tfhs
 
 RUN \
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 575159689BEFB442 && \
     apt-get update && \
     apt-get install -y \
         # Required by snappy-frames dependency.
--- a/ci_build/Dockerfile
+++ b/ci_build/Dockerfile
@@ -12,8 +12,6 @@ ADD . /tfhs
 WORKDIR /tfhs
 
 RUN \
-    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 575159689BEFB442 && \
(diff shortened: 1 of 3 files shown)


ci_build/Dockerfile
```diff
@@ -12,8 +12,6 @@ ADD . /tfhs
 WORKDIR /tfhs
 
 RUN \
-    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 575159689BEFB442 && \
```
blackgnezdo: This was slightly more end-to-end secure, but I guess we place ultimate trust into github anyway, so no big deal.

tensorflow-ops/tests/GradientTest.hs
```diff
@@ -392,18 +392,18 @@ transAttrs a b =
 testConv2DBackpropInputGrad :: Test
 testConv2DBackpropInputGrad = testCase "testConv2DBackpropInputGrad" $ do
     (dx, shapeDX, shapeX) <- TF.runSession $ do
-        let conv_input_shape = TF.vector [1, 2, 2, 1 :: Int32] -- [batch, h, w, in_channels]
```
blackgnezdo: Looks like an unrelated cleanup. Split it off?

ci_build/Dockerfile
fkm3: Whoops, got trigger happy. Restored it.

tensorflow-ops/tests/GradientTest.hs
fkm3: Reverted.

PR #192, opened 2018-05-19
Fix install script for High Sierra

mezzomondo:
Hi,

this is a small fix that allows a clean installation on OSX High Sierra/Sierra/El Capitan due to the limitations Apple introduced to write to /usr folder even for root.

### DIFF
--- a/tools/install_macos_dependencies.sh
+++ b/tools/install_macos_dependencies.sh
@@ -28,11 +28,12 @@ echo "Downloading libtensorflow..."
 curl https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.8.0.tar.gz > libtensorflow.tar.gz
 
 echo "Extracting and copying libtensorflow..."
-tar zxf libtensorflow.tar.gz -C .
-sudo rsync -a lib/ /usr/local/lib
-sudo rsync -a include/ /usr/local/include
+TMP_DIR=`mktemp -d`
+tar zxf libtensorflow.tar.gz -C $TMP_DIR
+sudo rsync -a $TMP_DIR/lib/ /usr/local/lib
+sudo rsync -a $TMP_DIR/include/ /usr/local/include
 rm libtensorflow.tar.gz
-rm -rf lib/ include/
+rm -rf $TMP_DIR
 sudo mv /usr/local/lib/libtensorflow.so /usr/local/lib/libtensorflow.dylib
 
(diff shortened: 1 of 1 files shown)


tools/install_macos_dependencies.sh
```diff
@@ -28,8 +28,11 @@ echo "Downloading libtensorflow..."
 curl https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.8.0.tar.gz > libtensorflow.tar.gz
 
 echo "Extracting and copying libtensorflow..."
-sudo tar zxf libtensorflow.tar.gz -C /usr/local
+tar zxf libtensorflow.tar.gz -C .
```
blackgnezdo: It is dangerous to presume you can write into the current directory. You can create a temp  directory in /temp with  mktemp -d.

blackgnezdo:
Thanks @mezzomondo for the fix. Once you fix up the minor nit, maybe @judah can confirm that it's working for him on Mac.

tools/install_macos_dependencies.sh
```diff
@@ -28,8 +28,12 @@ echo "Downloading libtensorflow..."
 curl https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.8.0.tar.gz > libtensorflow.tar.gz
 
 echo "Extracting and copying libtensorflow..."
-sudo tar zxf libtensorflow.tar.gz -C /usr/local
+TMP_DIR=`mktemp -d`
```
blackgnezdo: $() is nicer way to do this (allows effortless nesting unlike ``)

tools/install_macos_dependencies.sh
```diff
@@ -28,8 +28,12 @@ echo "Downloading libtensorflow..."
...
+tar zxf libtensorflow.tar.gz -C $TMP_DIR
```
blackgnezdo: I suspect mktemp is not out got get us, but "$TMP_DIR" would make any spaces it might throw our way harmless (please change all instances).

tools/install_macos_dependencies.sh
```diff
@@ -28,8 +28,12 @@ echo "Downloading libtensorflow..."
 curl https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.8.0.tar.gz > libtensorflow.tar.gz
```
blackgnezdo: Could you redirect this into TMP_DIR too?

PR #200, opened 2018-07-24
Simplified shell.nix

basvandijk:
* I upgraded nixpkgs to the [latest master which now contains `libtensorflow`](https://github.com/NixOS/nixpkgs/blob/7098bcac278a2d028036bb3a23508fd1c52155ac/pkgs/development/libraries/libtensorflow/default.nix). This replaces the custom `tensorflow-c`. Note that `libtensorflow` now also builds on OS X and has optional CUDA support.

* nixpkgs is now pinned using `builtins.fetchTarball` which means we can build with an empty `NIX_PATH` thus making the build pure.

  This does require Nix-2.0 but this has been released for a while with NixOS-18.03.

* The stack override has been replaced by passing `extraArgs = "--system-ghc";` to `buildStackProject`.

BTW I also added [overrides for the tensorflow-haskell packages to nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/haskell-modules/configuration-tensorflow-ghc-8.2.x.nix). This means that the following now builds:

```shell
for pkg in tensorflow tensorflow-proto tensorflow-core-ops tensorflow-logging tensorflow-mnist; do 
  nix-build -A haskell.packages.ghc822.$pkg; 
done
```

### DIFF
--- a/nix/src.json
+++ b/nix/src.json
@@ -1,6 +1,6 @@
 {
   "owner": "NixOS",
   "repo": "nixpkgs",
-  "rev": "729da53a147eda1a1678a313754fcda3a8791279",
-  "sha256": "17l19rbhb3wdrrcyjgq34bk3p658jbldzxp9q1scdvrvz0g99psb"
+  "rev": "7098bcac278a2d028036bb3a23508fd1c52155ac",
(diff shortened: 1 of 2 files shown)


basvandijk:
I signed it!

PR #201, opened 2018-07-25
Update to tensorflow 1.9

helq:
Mirroring changes of 1829c7e.

I've run `stack test` and it works without problems. I've tested the changes with tensorflow-haskell-deptyped and everything works just fine for me (though, I don't know if it works for everybody, my distribution (Archlinux) seems to package a slightly different version of tensorflow once in a while.)

PS: I really don't know what to put in the changelog. I thought you guys have already released version 0.2.0.0, so this would be just an update, but I'm not sure.


blackgnezdo:
Thanks for the PR. Will merge once the build clears.

PR #208, opened 2018-08-17
Document use of nvidia docker version 2

jcberentsen:


fkm3:
Thanks! Will merge after CI run completes.

PR #209, opened 2018-08-17
Haddock

blackgnezdo:
@mattludwigs, I had to massage the command a bit, but it worked out in the end.


mattludwigs:
@blackgnezdo Thank you for keeping me posted on this!

Issue #229, opened 2019-01-14
Question about the two types of variables


rikvdkleij:
To run the `minimizerWith` one has to use a `Variable a` but for saving and restoring the model parameters, one has to use a `Tensor Ref a`. It's possible to pass a `Variable a` to the save function (as a `Tensor Value a`) but then the saving of the parameters does not work.

What is the preferred solution to train the model, saving and restoring the model with respect to those two kind of variables?

rikvdkleij:
Finally found a solution by first creating `Variable` and then after the training creating  a Tensor Ref which is initialized by using the `readValue` of the variable.

Issue #228, opened 2019-01-14
Saving and Restoring the Session State (to Haskell)


schnecki:
Hi,

I was wondering whether it is possible to read the current session state, close the session, later open a new one and reload the former session state to be able to continue the calculations.

This would prevent of encapsulating the whole program into `SessionT`. I found the possibility of saving `Tensor`s, which however might not be the full state. 

Furthermore, I tried using `readerSerializeStateV2` and `readerRestoreState` but couldn't get anything running, nor I'm sure that these functions do what I'd like to do.

Has anyone done something like this yet and if so, is there a code example somewhere? Or might this not be possible at all yet?

rikvdkleij:
I have tried to find a solution based on https://github.com/tensorflow/haskell/blob/61e58fd33f50fdd70a34489a6d599118575d2068/tensorflow-mnist/tests/ParseTest.hs  which only contains the restoring part.

The problem is the saving part. It looks like applying `asGraphDef` on the node which contains the graph that it does not serializes everything because when I load the graph in the next session and then restore the model parameters (which have the same name as in the training) TF gives an error : 
`*** Exception: TensorFlowException TF_INVALID_ARGUMENT "Node 'Assign_7': Unknown input node 'b'"`
(`b` is name of model parameter)

A part of the code which is relevant for saving the graph:
```
import qualified TensorFlow.Variable as V

    w <- V.zeroInitializedVariable' (TF.opName .~ "w") (TF.Shape [])
    b <- V.zeroInitializedVariable' (TF.opName .~ "b") (TF.Shape [])
    inputs :: TF.Tensor TF.Value Float <- TF.build $  TF.placeholder'(TF.opName .~ "x")  [-1, fromIntegral (length xData)]
    let yHat = (inputs `TF.mul` V.readValue w) `TF.add` V.readValue b

    let graph = TF.asGraphDef $ TF.render yHat
    let graphMessage = encodeMessage graph
    liftIO $ BS.writeFile "graph.pb" graphMessage
```

Later on the trained model parameters (`w_v` and `b_v` are the values) are stored  because TF `Variable` can not be used for that.
```
  TF.runSession $ do
    w1 <- TF.initializedVariable' (TF.opName .~ "w") (TF.scalar w_v)
    w2 <- TF.initializedVariable' (TF.opName .~ "b") (TF.scalar b_v)
    let path = B8.pack "checkpoint.chkpt"
    TF.save path [w1, w2] >>= TF.run_
```

[...]

PR #232, opened 2019-02-05
Added support for minimizing on Tensor Ref besides Variable

rikvdkleij:


fkm3:
Thanks for the PR!

I'm thinking it would be cleaner to fork Minimize.hs, i.e. add a new module that only works for `Tensor Ref`, because:
* The code is fairly short, so the duplication isn't that bad
* Using `Tensor Ref` is the old way we'd like to get rid of at some point and that would make it easier to remove at a later time
* The new `Minimizer` type is a bit scary (this is true of the library in general though)
* `Tensor Ref` and `Variable` have subtly different semantic so it is difficult to verify the shared parts of the implementation (e.g. the runtime error you saw is because `Variable` guarantees that reads will happen after initialization while `Tensor Ref` does not)

What do you think?

rikvdkleij:
@fkm3 
Thanks for your review!

I would have been appreciated a lot if you would have responded to issues(#228 and #229)  related to this PR (to give some direction at least). It was really a struggle to get something (nice) working with saving and restoring model parameters. Still do have no solution for saving and restoring the model itself. There is no documentation and reading the code helps but in the end the interface to TF itself is not understandable for me.  You would expect from a machine learning library that saving and restoring model (parameters) would be trivial.

> I'm thinking it would be cleaner to fork Minimize.hs, i.e. add a new module that only works for `Tensor Ref`

I do not agree because it contains 2 encapsulated implementations of the minimizer which share the types and logic. Also not of course because I wrote it 😄 

> * The code is fairly short, so the duplication isn't that bad

That is not the case for `adam`. Logic is the same.

> * Using `Tensor Ref` is the old way we'd like to get rid of at some point and that would make it easier to remove at a later time

That is really something that I should have heard earlier... It is also really confusing for newcomers that there a two kinds of variables.

> * The new `Minimizer` type is a bit scary (this is true of the library in general though)

The exported functions are safe. The data constructor is not exposed. It's not less safe than rest of library.

> * `Tensor Ref` and `Variable` have subtly different semantic so it is difficult to verify the shared parts of the implementation (e.g. the runtime error you saw is because `Variable` guarantees that reads will happen after initialization while `Tensor Ref` does not)

Yes, that was a unexpected surprise which I found out after testing but that's now solved. 


Finally. I really hope this library is developed further and will become easier and safer to use because TF itself works great and Haskell is great language! Also with TF 2 in the pipeline.

fkm3:
Sorry for the slow response.

This library is a volunteer effort and obviously incomplete. Your are blazing the trail in this case :)

> > The code is fairly short, so the duplication isn't that bad
>
> That is not the case for `adam`. Logic is the same.

I'm not say there is no duplication, I'm claiming that "it isn't that bad". The implementation of adam is about 25 lines.

> > Using Tensor Ref is the old way we'd like to get rid of at some point and that would make it easier to remove at a later time
>
> That is really something that I should have heard earlier... It is also really confusing for newcomers that there a two kinds of variables.

Agreed. They are actually completely different implementations in libtensorflow. The newer one (called  resource handles in C++) was added after the initial design and implementation of this library. Right now, we are half migrated to the new one. I haven't looked into tensorflow 2.0 yet, but it may be that both are deprecated and there is something yet newer we need to migrate to.

> > The new Minimizer type is a bit scary (this is true of the library in general though)
>
> The exported functions of Minimize are safe. The data constructor is not exposed. It's not less safe than rest of library.

I don't mean that it is unsafe, I mean that it is unintuitive. Even as one of the maintainers of this library, I have to stare at the type for a while to figure out what arguments it can accept.

I think the data constructor (or type alias as it was before) should be exported so that users can define their own Minimizers and/or transform Minimizer (e.g. to apply gradient clipping).

blackgnezdo:
@rikvdkleij if you still care about this, could you check if our HEAD version with TF 2.3 support is any better for what you do? Feel free to create a new PR is it works for you.

rikvdkleij:
@blackgnezdo thanks for the heads up but I’m not involved anymore in the project which was using Haskell TF. It’s also too long ago to give proper answer.

PR #233, opened 2019-02-27
add setup/deployment script and description for CentOS

subbyte:
I am using the tensorflow Haskell binding on CentOS with stack only (no Docker or Nix). And I think it would be convenient to automatically setup the running environment when someone else uses it. So I put up a dependency solving script similar to the MacOS one and add description to `README.md`. The environment is set for stack resolver lts-12.26 with GHC 8.4.4 and everything is consistent with the latest version in this git repo. The script is tested via `stack test` with success. Feel free to let me know if I need to revise anything before the merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,21 @@ After running the script to install system dependencies, build the project with 
 
     stack test
 
+## Build on CentOS (native OS with stack)
+
+1. install `stack` with version >1.4.0
+
+2. run the [install_centos_dependencies.sh](./tools/install_centos_dependencies.sh)
+script in the `tools/` directory to install dependencies.
+- Protocol Buffers (git version to support the latest build with GHC 8.4.4)
+- Snappy (install via `yum`)
+- libtensorflow (deployed to `/usr/lib64/` and `/usr/include`)
+
+3. test with this library repo via `stack test`
+
+4. develop your own project with `resolver: lts-12.26` in the `stack.yaml` file
+   and an `extra-deps` entry pointed to this repo.
+
 ## Build on NixOS
 
 The `shell.nix` provides an environment containing the necessary
--- a/tools/install_centos_dependencies.sh
+++ b/tools/install_centos_dependencies.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+echo "Installing CentOS 7.6 Dependencies"
+echo "=================================="
+
+# Basic environment
+echo "Checking basic tools..."
+sudo yum install wget git -y
+
+# Protocol Buffers
+echo "Preparing compiling environment..."
+sudo yum install autoconf automake libtool unzip gcc-c++ -y
+TMP_DIR=$(mktemp -d)
+pushd $TMP_DIR
+echo "Downloading protobuf..." 
+git clone https://github.com/google/protobuf.git
+sudo rm -rf /usr/local/src/protobuf
+sudo mv protobuf /usr/local/src/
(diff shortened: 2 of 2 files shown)


tools/install_centos_dependencies.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
blackgnezdo: Is this a valuable indirection compared to /bin/bash?

README.md
```diff
@@ -124,6 +124,21 @@ After running the script to install system dependencies, build the project with
 
     stack test
 
+## Build on CentOS (native OS with stack)
+
+1. install `stack` with version >1.4.0
```
blackgnezdo: Why not include this step into the script itself?

README.md
```diff
@@ -124,6 +124,21 @@ After running the script to install system dependencies, build the project with
...
+
+2. run the [install_centos_dependencies.sh](./tools/install_centos_dependencies.sh)
+script in the `tools/` directory to install dependencies.
```
blackgnezdo: Let's rephrase this in a shorter way:

2. run [tools/install_centos_dependencies.sh](./tools/install_centos_dependencies.sh)
to install dependencies.

tools/install_centos_dependencies.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo "Installing CentOS 7.6 Dependencies"
+echo "=================================="
+
+# Basic environment
+echo "Checking basic tools..."
+sudo yum install wget git -y
```
blackgnezdo: Is there a reason to not bundle up all yum invocations?

tools/install_centos_dependencies.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Protocol Buffers
+echo "Preparing compiling environment..."
+sudo yum install autoconf automake libtool unzip gcc-c++ -y
```
blackgnezdo: I'm wary of the scripts that run ask for my password. Just make the user run `sudo` explicitly when they invoke the script. If you must drop privileges for some operations like network access with git, you can reduce privileges inside the script ($SUDO_USER remains set).

subbyte:
> Thanks for your contribution!
> 
> I started reviewing this script and collected a few comments. I realized we shouldn't be keeping (much less maintaining) recipes for multiple OSs. I didn't think it through when we merged Nix.
> 
> Would mind hosting the script in a separate repo? I'm happy to include a link to it into our README.

Good suggestion! Please refer to the two links for setting it up on CentOS. You can post both or only the first link, which has a link to the second in itself.
- [Instruction Description](https://github.com/subbyte/haskell-learn/blob/master/tensorflow_setup.md)
- [Env Setup Script](https://github.com/subbyte/haskell-learn/blob/master/scripts/install_tensorflow_deps_centos.sh)

I also edit the script following your comments. Thanks for the feedback!
- I keep `#!/usr/bin/env bash` for generality, in case people edit the script and use it on other systems.
- I put a stack installation section into the script, and it setup the complete runtime now.
- I do not bundle the yum installations since people may comment out a section and run another.
- I keep the `sudo` in the current style to keep consistent style with the stack installation script.

Please drop this pull request after you put the links in the README file.

blackgnezdo:
Thanks: https://github.com/tensorflow/haskell/commit/896a0d31f75a74748879f7338bb44aac376e20c6

PR #236, opened 2019-03-12
Add gradient for sqrt function

erikabor:
I added the gradient for the `sqrt` function. Please check if the implementation is right.

### DIFF
--- a/tensorflow-ops/src/TensorFlow/Gradient.hs
+++ b/tensorflow-ops/src/TensorFlow/Gradient.hs
@@ -813,6 +813,10 @@ opGrad "Placeholder" _ _ _ = []
 opGrad "VarHandleOp" _ _ _ = []
 opGrad "Variable" _ _ _ = []
 
+opGrad "Sqrt" _ [toT -> x] [dz] = [Just $ sq' `CoreOps.mul` dz]
+  where
+    sq' = vector [1] `CoreOps.div` (vector [2] `CoreOps.mul` CoreOps.sqrt x)
+
 opGrad n nodeDef ins grads =
     error $ "no gradient implemented for " ++
             show (n, length ins, length grads, showMessage nodeDef, ins)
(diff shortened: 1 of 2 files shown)


tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -813,6 +813,10 @@ opGrad "Placeholder" _ _ _ = []
 opGrad "VarHandleOp" _ _ _ = []
 opGrad "Variable" _ _ _ = []
 
+opGrad "Sqrt" _ [toT -> x] [dz] = [Just $ sq' `CoreOps.mul` dz]
+  where
+    sq' = vector [1] `CoreOps.div` (vector [2] `CoreOps.mul` CoreOps.sqrt x)
```
blackgnezdo: I don't remember, is there a way to multiply by a scalar rather than a (implicitly?) replicated vector?

fkm3: I think

```haskell
scalar 1 `CoreOps.div` (scalar 2 `CoreOps.mul` CoreOps.sqrt x)
```

would work,  but if that doesn't compile, then the proposed implementation is fine.

tensorflow-ops/tests/GradientTest.hs
```diff
@@ -324,6 +324,14 @@ testPad =
     V.fromList [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1] @=? dx
     V.fromList [2, 2, 3] @=? s
 
+testSqrt :: Test
+testSqrt = testCase "testSqrt" $ do
+    [dx] <- TF.runSession $ do
+        x <- TF.render $ TF.vector [0.25 :: Float]
```
fkm3: Could you choose an input where the expected gradient isn't 1 or 0?

tensorflow-ops/src/TensorFlow/Gradient.hs
erikabor: Yes, the multiplication with a scalar works. It compiles.

tensorflow-ops/tests/GradientTest.hs
erikabor: Yes, I've just fixed it.

PR #237, opened 2019-03-19
update to ghc 8.6.4

leptonyu:


leptonyu:
I signed it!

leptonyu:
Upgrade to ghc864 has some issues to be solved.

1. `snappy-framing` is not designed for CRC32C. It's a heavy dependency, and it depends on `snappy`, which is broken when upgrade ghc864 using nix. I tries install `snappy` without nix, and succeed. So I should keep this unchanged, Just add `extra-deps` snappy-0.2.0.2 to stack.yaml
2. `proto-lens` has upgrade to 0.4.*, which remove `proto-lens-protoc` package, split into `proto-lens-runtime` and `proto-lens-setup`. see  [Proto-lens ReadMe](https://github.com/google/proto-lens#readme).
3. `proto-lens-0.4.*` remove `Default` instance, see [Don't use `data-default-class`](https://github.com/google/proto-lens/pull/210). A lot code change to replace `def` as `defMessage`.
4. ghc8.6 enabled [MonadFailDesugaringbydefault](https://gitlab.haskell.org/ghc/ghc/wikis/migration/8.6#MonadFailDesugaringbydefault), which cause no instance of `MonadFail Identity` error. This one can be resolved by specifying extension `NoMonadFailDesugaring`, so I will revert related code. `BuildT` and `SessionT` add derived `MonadFail`  instance for more convenient with this issue.

leptonyu:
5. One more issue, but should be unchanged in this PR. warning with `sDistHook`, this hook was deprecated. But I will remove the changed.

fkm3:
If you can find a way to get rid of the snappy dependency, that would be very appreciated. I agree it should go into another PR though.

leptonyu:
I searched hackage.haskell.org, it seems that there is no crc32c package available. One way is asking the author of `snappy-framing` split the code to publish a new package for crc32c. Another way is to publish a binding package to official [crc32c](https://github.com/google/crc32c) lib, which I can do it.

Then `snappy-framing` can be removed.

leptonyu:
[haskell-crc32c](https://github.com/leptonyu/crc32c) now can work by replacing `snappy-framing` in this project. There is no crc32c package in nix, I think this might be a problem for new users.

blackgnezdo:
Thanks for extracting a smaller dependency!

Does it make sense to publish this package to hackage while you are at it?

leptonyu:
@blackgnezdo Yes, if you want to use this package, I will publish it to hackage.

What's wrong with Kokoro build failed? It seems to that I cannot see the result.

leptonyu:
I have updated the code for [haskell-crc32c](https://github.com/leptonyu/crc32c), including cpp source in it. Then this package can be built without any dependency of libcrc32c.

fkm3:
There are some build warnings and we use the `--pedantic` flag when building the tests, so it fails on the warnings.

Here are the current warnings:

```
    /tfhs/tensorflow/src/TensorFlow/Types.hs:70:1: error: [-Wunused-imports, -Werror=unused-imports]
        The import of `Data.Default' is redundant
          except perhaps to import instances from `Data.Default'
        To import instances alone, use: import Data.Default()
       |
    70 | import Data.Default (def)
       | ^^^^^^^^^^^^^^^^^^^^^^^^^
    
    /tfhs/tensorflow/src/TensorFlow/Types.hs:91:1: error: [-Wdodgy-imports, -Werror=dodgy-imports]
        The import item `AttrValue(..)' suggests that
        `AttrValue' has (in-scope) constructors or class methods,
        but it has none
       |
    91 | import Proto.Tensorflow.Core.Framework.AttrValue
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
    
    /tfhs/tensorflow/src/TensorFlow/Types.hs:91:1: error: [-Wdodgy-imports, -Werror=dodgy-imports]
        The import item `AttrValue'ListValue(..)' suggests that
        `AttrValue'ListValue' has (in-scope) constructors or class methods,
        but it has none
       |
    91 | import Proto.Tensorflow.Core.Framework.AttrValue
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
    
    /tfhs/tensorflow/src/TensorFlow/Types.hs:108:1: error: [-Wdodgy-imports, -Werror=dodgy-imports]
        The import item `TensorProto(..)' suggests that
        `TensorProto' has (in-scope) constructors or class methods,
        but it has none
        |
    108 | import Proto.Tensorflow.Core.Framework.Tensor as Tensor
        | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
    
    /tfhs/tensorflow/src/TensorFlow/Types.hs:122:1: error: [-Wdodgy-imports, -Werror=dodgy-imports]
        The import item `TensorShapeProto(..)' suggests that
        `TensorShapeProto' has (in-scope) constructors or class methods,
        but it has none
        |
    122 | import Proto.Tensorflow.Core.Framework.TensorShape
        | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

fkm3:
In case it is helpful: The script that runs the tests in haskell/ci_build/outer_launch_tests.sh

leptonyu:
@fkm3 i've fixed the warnings.

fkm3:
Looks good to me, but I'll wait for @blackgnezdo to approve.

blackgnezdo:
Appreciate the upgrade! Thank you!

PR #238, opened 2019-03-30
Avoid computing gradients for incidental nodes

jcberentsen:
This implements a TODO by fmayle

The `testGradientIncidental` has an incidental "Diag" node, which has no `opGrad` implementation. Since the Diag node is not a successor of the 'x' we want gradient for, it won't be visited.

I removed two `opGrad` implementations as suggested by comments by fmayle. This may be a breaking change, and I'm open to putting these back if needed.

### DIFF
--- a/tensorflow-ops/src/TensorFlow/Gradient.hs
+++ b/tensorflow-ops/src/TensorFlow/Gradient.hs
@@ -36,6 +36,7 @@ import Data.Int (Int32, Int64)
 import Data.Foldable (foldlM)
 import Data.List (foldl', sortBy)
 import Data.Map.Strict (Map)
+import qualified Data.IntSet as IntSet
 import Data.Maybe (fromMaybe, maybeToList, mapMaybe)
 import Data.Ord (comparing)
 import Data.ProtoLens.TextFormat (showMessage)
@@ -165,6 +166,11 @@ gradients y xs = build $ do
         (\f x -> fromMaybe (error $ "no NodeDef found for " ++ show x) (f x))
         . flip Map.lookup
     let (gr, nodeMap) = createGraph yName nodeDefLookup
+        xnodes = mapMaybe (\x -> nodeMap ^. (at . outputNodeName . renderedOutput $ x)) xs
+        -- make a set of the nodes reachable from the xnodes
+        -- The xnodes are not part of this set (unless reachable from another xnode)
+        reachableSet = computeReachableSet xnodes gr
+
     -- Set gradient of y to one.
     -- TODO: nicer
     let initPending :: Map.Map FGL.Node (PendingGradients a)
@@ -175,13 +181,20 @@ gradients y xs = build $ do
                                 .~ [yOne]
                                 )
     -- Calculate the gradients of y w.r.t. each node in the graph.
-    gradientMap <- graphGrads gr initPending
+    gradientMap <- graphGrads gr reachableSet initPending
     -- Lookup the gradients for each x.
     forM xs $ \x ->
         let Output i xName = renderedOutput x
         in maybe (render $ zerosLike $ toTensor x) return $ do
             n <- nodeMap ^. at xName
             gradientMap ^. at n . nonEmpty . outputIxAt i
+
+-- | Compute a set of nodes reachable from the start nodes
+--
+-- the start nodes are excluded, unless reachable from another start node
+computeReachableSet :: [FGL.Node] -> Graph -> IntSet.IntSet
+computeReachableSet vs g =
+  IntSet.fromList $ concatMap (drop 1 . FGL.preorder) (FGL.dff vs g)
 
 outputIxAt :: OutputIx -> Lens' (IntMap.IntMap v) (Maybe v)
 outputIxAt = intAt . unOutputIx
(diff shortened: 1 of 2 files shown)


tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -245,16 +253,17 @@ nonEmpty = anon mempty null
...
-graphGrads gr initPending = view gradientsResult <$> foldlM go initState nodeOrder
+graphGrads gr reachableSet initPending = view gradientsResult <$> foldlM go initState nodeOrder
   where
     initState = GradientsState initPending Map.empty
     -- Reverse topological sort.
-    -- TODO(fmayle): Filter out nodes that are not successors of any x in xs to
+    -- Filter out nodes that are not successors of any x in xs to
     -- avoid calculating gradients that won't be used.
-    nodeOrder = FGL.topsort $ FGL.grev gr
+    nodeOrder = filter (`IntSet.member` reachableSet) . FGL.topsort . FGL.grev $ gr
```
fkm3: I think there is a bit of an off by one error. If `xnodes` are in `reachableSet`, then we'll try to compute the gradients of the inputs to the nodes that output the `xs` tensors. That is probably why you couldn't delete the gradient functions for the rest of the undifferentiable ops.

One idea is to remove `xnodes` from `reachableSet` and then make the make the `reachableSet` membership check inside of `go`, after finalizing `outputGrads`, but before computing `inputGrads`.

```haskell
    nodeOrder = FGL.topsort $ FGL.grev gr
    go :: GradientsState a -> Int -> Build (GradientsState a)
    go state node = do
        -- Aggregate the accumulated gradients for this node.
        outputGrads <-
                sumPendingGradient (state ^. gradientsPending . at node . nonEmpty)
        if null outputGrads
           then pure state
           else do
              let nextState = state & gradientsResult %~ Map.insert node outputGrads
              if node `IntSet.member` reachableSet
                then do
                  let ctx = FGL.context gr node
                  inputGrads <- calculateInputGrads ctx outputGrads gr
                  -- Calculate the gradients for each of the node's inputs.
                  pure $ updatePendingGradients ctx inputGrads nextState
                else
                  pure nextState
```

(sorry the suggested code is ugly, my haskell skills are rusting)

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -165,6 +166,10 @@ gradients y xs = build $ do
         (\f x -> fromMaybe (error $ "no NodeDef found for " ++ show x) (f x))
         . flip Map.lookup
     let (gr, nodeMap) = createGraph yName nodeDefLookup
+        xnodes = mapMaybe (\x -> nodeMap ^. (at . outputNodeName . renderedOutput $ x)) xs
+        -- http://hackage.haskell.org/package/fgl-5.7.0.1/docs/Data-Graph-Inductive-Graph.html#g:9
```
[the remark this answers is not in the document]
jcberentsen: I'll remove it. (It belonged with `reachables`, referring to some fgl traversal code)

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -175,14 +180,17 @@ gradients y xs = build $ do
...
-    gradientMap <- graphGrads gr initPending
+    gradientMap <- graphGrads gr reachableSet initPending
     -- Lookup the gradients for each x.
     forM xs $ \x ->
         let Output i xName = renderedOutput x
         in maybe (render $ zerosLike $ toTensor x) return $ do
             n <- nodeMap ^. at xName
             gradientMap ^. at n . nonEmpty . outputIxAt i
 
+reachables :: [FGL.Node] -> Graph -> [FGL.Node]
```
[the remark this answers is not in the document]
jcberentsen: You're correct, the duplicates go away in the `IntSet`. I'll refactor into a `computeReachableSet` function.

tensorflow-ops/src/TensorFlow/Gradient.hs
jcberentsen: Thank you for the feedback and suggestions! This has been implemented and the `testEmbeddingLookupGradients` is fixed.

tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -165,6 +166,10 @@ gradients y xs = build $ do
         (\f x -> fromMaybe (error $ "no NodeDef found for " ++ show x) (f x))
         . flip Map.lookup
     let (gr, nodeMap) = createGraph yName nodeDefLookup
+        xnodes = mapMaybe (\x -> nodeMap ^. (at . outputNodeName . renderedOutput $ x)) xs
+        -- make a set of the nodes reachable from the xnodes (lacking the xnodes themselves)
+        reachableSet = (computeReachableSet xnodes gr) `IntSet.difference` (IntSet.fromList xnodes)
```
fkm3: Sorry, I missed an edge case when I made this suggestion. If one member of `xnodes` is reachable from another member of `xnodes`, then we need to keep it in this set.

Maybe one solution is to remove the root of each `Tree` returned by the `FGL.dff` call before calling `FGL.preorderF`.

jcberentsen: Good catch! A test for this case and your suggestion has been implemented.

fkm3:
Thanks!

PR #239, opened 2019-04-02
Add gradient for ResizeBilinear

jcberentsen:


tensorflow-ops/src/TensorFlow/Gradient.hs
```diff
@@ -817,6 +817,18 @@ opGrad "Tile" _ [toT -> x, toT -> multiples] [dz] =
     axes = CoreOps.range 0 (CoreOps.size splitShape) (2 :: Tensor Build Int32)
     reshapedDz = CoreOps.reshape dz splitShape
 
+opGrad "ResizeBilinear" nodeDef [toT -> x, _] [dz] =
+    [ Just $ CoreOps.cast $ CoreOps.resizeBilinear'
```
fkm3: The Python implemention is different:
https://github.com/tensorflow/tensorflow/blob/7a056bf82591cd92a0e918e444b2e44e076840c1/tensorflow/python/ops/image_grad.py#L52

Should we use `CoreOps.resizeBilinearGrad'` or is there a reason to prefer this implementation?

jcberentsen: You're right! For some reason I didn't spot that one! I'll change the code to use `CoreOps.resizeBilinearGrad'`instead!

jcberentsen: Using `CoreOps.resizeBilinearGrad'` also exposed my initial attempt was wrong. Now the gradients in the test are in correspondence with `tile`.

fkm3:
Thanks!

PR #241, opened 2019-04-10
Make 'mean' doubly differentiable

jcberentsen:
Use stopGradient on shape computations
Add opGrad for StopGradient


fkm3:
Thanks!

PR #248, opened 2019-09-07
Switch to LTS 14 and bump versions to 0.2.0.1

blackgnezdo:


judah:
Just checking: is it intentional that you're not changing to the packages other than bumping their versions?  Do you have specific plans for releasing new versions?

blackgnezdo:
> Just checking: is it intentional that you're not changing to the packages other than bumping their versions? Do you have specific plans for releasing new versions?

Yeah, I was planning to release this to Haddock under a bumped version.

Issue #249, opened 2019-09-08
Fix deprecated API use "SDist hooks violate the invariants of new-sdist"


blackgnezdo:
tensorflow-core-ops        > configure (lib)                                                                                                                                                                        
tensorflow-core-ops        > [1 of 2] Compiling Main             ( /tfhs/tensorflow-core-ops/Setup.hs, /tfhs/tensorflow-core-ops/.stack-work/dist/x86_64-linux/Cabal-2.4.0.1/setup/Main.o )                         
tensorflow-core-ops        >                                                                 
tensorflow-core-ops        > /tfhs/tensorflow-core-ops/Setup.hs:53:7: warning: [-Wdeprecations]                                                                                                                     
tensorflow-core-ops        >     In the use of ‘sDistHook’                                                                                                                                                          
tensorflow-core-ops        >     (imported from Distribution.Simple, but defined in Distribution.Simple.UserHooks):                                                                                                 
tensorflow-core-ops        >     Deprecated: "SDist hooks violate the invariants of new-sdist. Use 'autogen-modules' and 'build-tool-depends' instead."                                                             
tensorflow-core-ops        >    |                                                                                                                                                                                   
tensorflow-core-ops        > 53 |     , sDistHook = \p maybe_l h f -> case maybe_l of                                                                                                                               
tensorflow-core-ops        >    |       ^^^^^^^^^                                                                                                                                                                   
tensorflow-core-ops        >                                                                                                                                                                                        
tensorflow-core-ops        > /tfhs/tensorflow-core-ops/Setup.hs:57:21: warning: [-Wdeprecations]                                                                                                                     
tensorflow-core-ops        >     In the use of ‘sDistHook’                                                                                                                                                           
tensorflow-core-ops        >     (imported from Distribution.Simple, but defined in Distribution.Simple.UserHooks):                                                                                                 
tensorflow-core-ops        >     Deprecated: "SDist hooks violate the invariants of new-sdist. Use 'autogen-modules' and 'build-tool-depends' instead."                                                             
tensorflow-core-ops        >    |                                                                                                                                                                                   
tensorflow-core-ops        > 57 |                     sDistHook hooks (fudgePackageDesc l p) maybe_l h f
tensorflow-core-ops        >    |                     ^^^^^^^^^                                                                                                                                                     
[...]

PR #250, opened 2019-09-08
Add autogen-module to unblock upload to hackage (related to #249)

blackgnezdo:


blackgnezdo:
This is the commit which became v0.2.0.1 for tensorflow-proto. I will tag the tree as such.

PR #254, opened 2020-02-14
nix updated to be compatible with LTS14.4

saeedhadikhanloo:
With @curiousleo, we have noticed some outdated versions in installation by nix. 
We did the following changes:
* `nix/update` updated to pull from the nix release `19.09`
* `shell.nix` updated to use `ghc865` as default for `ghc` (since `LTS14.4` uses `ghc865`).

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -11,8 +11,8 @@ let
   pkgs = import nixpkgs {};
 in
   pkgs.haskell.lib.buildStackProject {
-    # Either use specified GHC or use GHC 8.6.4 (which we need for LTS 13.13)
-    ghc = if isNull ghc then pkgs.haskell.compiler.ghc864 else ghc;
+    # Either use specified GHC or use GHC 8.6.5 (which we need for LTS 14.4)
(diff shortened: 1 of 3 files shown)


saeedhadikhanloo:
@googlebot I signed it!

shell.nix
```diff
@@ -11,8 +11,8 @@ let
   pkgs = import nixpkgs {};
 in
   pkgs.haskell.lib.buildStackProject {
-    # Either use specified GHC or use GHC 8.6.4 (which we need for LTS 13.13)
-    ghc = if isNull ghc then pkgs.haskell.compiler.ghc864 else ghc;
+    # Either use specified GHC or use GHC 8.6.5 (which we need for LTS 14.4)
```
curiousleo: (Brings this file in line with https://github.com/tensorflow/haskell/pull/248)

blackgnezdo:
I don't use nor understand Nix, but since you guys think it's good, I'll merge.

PR #256, opened 2020-04-08
Unbreak build with Cabal >= 2.4.

mikesperber:
Just conditionally remove sDistHook in that case.


blackgnezdo:
This looks good to me, would love a second pair of eyes if @judah has cycles.

judah:
I checked what proto-lens is doing. We removed the sdist hook altogether, and instead just set the [autogen-modules] field (requires Cabal >=2.0), which allows you to specify a generated module without uploading a "dummy" version to Hackage.  See for example:
https://github.com/google/proto-lens#using-in-a-cabal-or-stack-package

@mikesperber there's not any reason to support older Cabal, could we do the same thing here (delete the sdist hook altogether and use autogen-modules in tensorflow-core-ops)? Otherwise, I think you won't be able to upload this package to Hackage anymore when using a newer Cabal version.

mikesperber:
Done.  Thanks for the review!

PR #268, opened 2020-11-10
Update Package Version Numbers

jcmartin:
This pull request simply updates the package version numbers following the Haskell Package Version Policy: https://pvp.haskell.org/ .

If we agree on the version numbers, the next pull request will be for an update haddock documentation, which is presently out of date.


tensorflow-opgen/tensorflow-opgen.cabal
```diff
@@ -17,7 +17,11 @@ library
   exposed-modules:     TensorFlow.OpGen.ParsedOp
                      , TensorFlow.OpGen
   build-depends:  proto-lens == 0.7.*
+<<<<<<< HEAD
```
blackgnezdo, line 20: Oops.

blackgnezdo (changes requested):
Could you fix up the conflict marker?

tensorflow-opgen/tensorflow-opgen.cabal
jcmartin: Thanks. Done.

tensorflow-logging/tensorflow-logging.cabal
```diff
@@ -29,11 +29,11 @@ library
                 , stm
                 , stm-chans
                 , stm-conduit
-                , tensorflow == 0.2.*
-                , tensorflow-core-ops == 0.2.*
-                , tensorflow-ops == 0.2.*
-                , tensorflow-proto == 0.2.*
-                , tensorflow-records-conduit == 0.1.*
+                , tensorflow >= 0.2.0
```
blackgnezdo, line 32: I don't think we tested the combinations, so it's probably safer to just change all the versions together and keep them at the same 0.3 value. WDYT?

blackgnezdo (changes requested):

tensorflow-logging/tensorflow-logging.cabal
jcmartin: Since the haskell files (*.hs) were not altered when I patched for TF 2.3, it is safe to say that they are compatible with both the old and new version. However, the `proto-lens` dependency update (in order to move to a newer version of LTS) did cause a problem in terms of dependency conflict.

Doing this ensures that if someone wanted to patch an older version of the libraries with the new proto-lens package, they would be able to. However, I will admit that this is a pretty fringe case.

Thoughts?

blackgnezdo: It's probably inconsequential either way. I'll merge as is.

blackgnezdo (approved):

PR #269, opened 2020-11-12
Update to haddock files for tensorflow-0.3 package (TensorFlow 2.3.0).

jcmartin:
As promised, this pull request completes the last thing remaining (excluding feature requests) for TensorFlow 2.3.0: the haddock files. There was also a minor edit to refer to 2.3.0-gpu and not 1.14.0-gpu in the README.

It would be very nice if those with the right permissions could push this upstream to Hackage.

For someone who has the spare cycles, it would be nice to push this up into Stackage as well:
https://github.com/commercialhaskell/stackage/blob/master/MAINTAINERS.md#adding-a-package

At this point in time, to add to Stackage, one must ensure compatibility with GHC 8.10.2 since that is what version Nightly is currently using. Special attention should be given to editing `debian-bootstrap.sh` to ensure that `libtensorflow-2.3.0` and `protoc-3.13.0` is properly installed along with any other non-Haskell dependencies (see `ci_build/Dockerfile` for the complete list).


blackgnezdo (changes requested):
I there are some excess files, e.g. we have both 
A docs/haddock/tensorflow-0.2.0.1/TensorFlow-Build.html (0)
...
A docs/haddock/tensorflow-0.3.0.0/TensorFlow-Build.html (0)

Maybe you meant to keep only the latter directory?

jcmartin:
Good catch. I had some issues with `tools/haddock.sh`, and it seems I didn't fully resolve it.

blackgnezdo (approved):

PR #281, opened 2022-12-22
Fix for GHC 9.2

Lucus16:
This fixes the package for GHC 9.2 and its corresponding bytestring version which has removed the `Data.ByteString.Lazy.Builder` module. It has been deprecated for a while so I expect most projects will have upgraded to a sufficiently recent version by now.


mihaimaruseac (changes requested):
Please sign CLA

Lucus16:
No thanks. The rest of the free software world works just fine without CLAs.

PR #285, opened 2023-01-17
Migrate from TF_DeprecatedSession to TF_Session

Minnozz:
This is a prerequisite for supporting SavedModel.

### DIFF
--- a/tensorflow/src/TensorFlow/Internal/FFI.hs
+++ b/tensorflow/src/TensorFlow/Internal/FFI.hs
@@ -20,8 +20,15 @@ module TensorFlow.Internal.FFI
     ( TensorFlowException(..)
     , Raw.Session
     , withSession
-    , extendGraph
     , run
+
+    , SessionAction
+
+    , Raw.SessionOptions
+
+    , Raw.Graph
+    , extendGraph
+
     , TensorData(..)
     , setSessionConfig
     , setSessionTarget
@@ -40,16 +47,17 @@ import Control.Monad.Catch (MonadMask, Exception, throwM, bracket, finally, mask
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Bits (Bits, toIntegralSized)
 import Data.Int (Int64)
+import Data.Foldable (for_)
 import Data.Maybe (fromMaybe)
 import Data.Typeable (Typeable)
 import Data.Word (Word8)
-import Foreign (Ptr, FunPtr, nullPtr, castPtr)
-import Foreign.C.String (CString)
-import Foreign.ForeignPtr (newForeignPtr, newForeignPtr_, withForeignPtr)
+import Foreign (Ptr, FunPtr, nullPtr, castPtr, with)
+import Foreign.ForeignPtr (newForeignPtr_)
 import Foreign.Marshal.Alloc (free)
 import Foreign.Marshal.Array (withArrayLen, peekArray, mallocArray, copyArray)
 import System.IO.Unsafe (unsafePerformIO)
 import qualified Data.ByteString as B
+import qualified Data.ByteString.Char8 as C
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.Encoding.Error as T
@@ -87,15 +95,26 @@ data TensorData = TensorData
     }
   deriving (Show, Eq)
 
+-- | The action can spawn concurrent tasks which will be canceled before
+-- withSession returns.
+type SessionAction m a = (IO () -> IO ()) -> Raw.Session -> Raw.Graph -> m a
+
 -- | Runs the given action after creating a session with options
 -- populated by the given optionSetter.
 withSession :: (MonadIO m, MonadMask m)
             => (Raw.SessionOptions -> IO ())
-            -> ((IO () -> IO ()) -> Raw.Session -> m a)
-            -- ^ The action can spawn concurrent tasks which will
-            -- be canceled before withSession returns.
+            -> SessionAction m a
             -> m a
-withSession optionSetter action = do
+withSession = withSession_ Raw.newSession
+
+withSession_ :: (MonadIO m, MonadMask m)
+             => (Raw.Graph -> Raw.SessionOptions -> Raw.Status -> IO Raw.Session)
+             -- ^ mkSession
+             -> (Raw.SessionOptions -> IO ())
+             -- ^ optionSetter
+             -> SessionAction m a
+             -> m a
+withSession_ mkSession optionSetter action = do
     drain <- liftIO $ newMVar []
     let cleanup s =
         -- Closes the session to nudge the pending run calls to fail and exit.
@@ -105,11 +124,12 @@ withSession optionSetter action = do
                 mapM_ shutDownRunner runners
                 checkStatus (Raw.deleteSession s)
     let bracketIO x y = bracket (liftIO x) (liftIO . y)
-    bracketIO Raw.newSessionOptions Raw.deleteSessionOptions $ \options -> do
-        bracketIO
-            (optionSetter options >> checkStatus (Raw.newSession options))
-            cleanup
-            (action (asyncCollector drain))
+    bracketIO Raw.newGraph Raw.deleteGraph $ \graph ->
+        bracketIO Raw.newSessionOptions Raw.deleteSessionOptions $ \options -> do
+            bracketIO
+                (optionSetter options >> checkStatus (mkSession graph options))
+                cleanup
+                (\session -> action (asyncCollector drain) session graph)
 
 asyncCollector :: MVar [Async ()] -> IO () -> IO ()
 asyncCollector drain runner = modifyMVarMasked_ drain launchAndRecord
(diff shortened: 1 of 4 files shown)


Minnozz:
With my very limited understanding of TensorFlow, I suspect that the old call to `TF_ExtendGraph` merges the supplied `GraphDef` with the current Session, while the new call to `TF_GraphImportGraphDef` does not consider the existing graph and therefore cannot resolve references to existing nodes.

Minnozz:
This is probably where the errors are coming from:
https://github.com/tensorflow/tensorflow/blob/fdfc646704c37bdf450525f6ced9d80df86e4993/tensorflow/core/common_runtime/graph_constructor.cc#L726-L736

The "old" ExtendGraph call ended up here:
https://github.com/tensorflow/tensorflow/blob/fdfc646704c37bdf450525f6ced9d80df86e4993/tensorflow/core/common_runtime/graph_execution_state.cc#L181-L286

Minnozz:
I will try using TF_NewOperation calls instead of using a GraphDef as intermediate.

fkm3:
That might be a lot of work. If I remember correctly, a lot of the haskell code was designed exploiting the fact that we could build up a graph of protos (without having to interact with the C API or do anything "impure") and then just chuck it all into tensorflow at once via `TF_ExtendGraph`, so a lot of refactoring might be necessary (I'd be happy to be wrong though).

Not sure if it is feasible, but maybe at the point where `TF_ExtendGraph` is currently called, instead we could traverse the `GraphDef` proto in haskell, translating its contents into calls to `TF_NewOperation` and friends. Of course it seems like a waste to build up the `GraphDef` and then only use it for that, but it might be the shortest path to a solution.

Minnozz:
Thanks for the feedback; I'll give it a try to see if it's feasible.

I'm currently using the `NodeDef` protos in the `Builder` before they are put into the `GraphDef` proto, but I think that is equivalent to your suggestion.

Minnozz:
It is indeed a lot of work. I will try using the TF_GraphImportGraphDef approach again, but this time with an input mapping for existing nodes in the graph.

Minnozz:
Passing an input map seems to work.

Minnozz:
@fkm3 Thanks for looking into this PR; can you tell something about the CI failure?

fkm3:
It is just some warnings-as-errors, you should be able to reproduce using the `--pedantic` flag of stack or ghc (or use the CI docker script).

```
tensorflow                 > [ 1 of 12] Compiling TensorFlow.Internal.Raw
tensorflow                 > 
tensorflow                 > /[tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs:86](https://cs.corp.google.com/piper///depot/google3/tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs?l=86):29: error: [-Wname-shadowing, -Werror=name-shadowing]
tensorflow                 >     This binding for ‘ptr’ shadows the existing binding
tensorflow                 >       bound at [src/TensorFlow/Internal/Raw.chs:86](https://cs.corp.google.com/piper///depot/google3/src/TensorFlow/Internal/Raw.chs?l=86):10
tensorflow                 >    |
tensorflow                 > 86 |     peek ptr = Output <$> {# get TF_Output->oper #} ptr
tensorflow                 >    |                             ^^^
tensorflow                 > 
tensorflow                 > /[tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs:87](https://cs.corp.google.com/piper///depot/google3/tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs?l=87):47: error: [-Wname-shadowing, -Werror=name-shadowing]
tensorflow                 >     This binding for ‘ptr’ shadows the existing binding
tensorflow                 >       bound at [src/TensorFlow/Internal/Raw.chs:86](https://cs.corp.google.com/piper///depot/google3/src/TensorFlow/Internal/Raw.chs?l=86):10
tensorflow                 >    |
tensorflow                 > 87 |                       <*> (fromIntegral <$> {# get TF_Output->index #} ptr)
tensorflow                 >    |                                               ^^^
tensorflow                 > 
tensorflow                 > /[tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs:89](https://cs.corp.google.com/piper///depot/google3/tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs?l=89):11: error: [-Wname-shadowing, -Werror=name-shadowing]
tensorflow                 >     This binding for ‘ptr’ shadows the existing binding
tensorflow                 >       bound at [src/TensorFlow/Internal/Raw.chs:88](https://cs.corp.google.com/piper///depot/google3/src/TensorFlow/Internal/Raw.chs?l=88):10
tensorflow                 >    |
tensorflow                 > 89 |         {# set TF_Output->oper #} ptr oper
tensorflow                 >    |           ^^^
tensorflow                 > 
tensorflow                 > /[tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs:90](https://cs.corp.google.com/piper///depot/google3/tfhs/tensorflow/src/TensorFlow/Internal/Raw.chs?l=90):11: error: [-Wname-shadowing, -Werror=name-shadowing]
tensorflow                 >     This binding for ‘ptr’ shadows the existing binding
tensorflow                 >       bound at [src/TensorFlow/Internal/Raw.chs:88](https://cs.corp.google.com/piper///depot/google3/src/TensorFlow/Internal/Raw.chs?l=88):10
tensorflow                 >    |
tensorflow                 > 90 |         {# set TF_Output->index #} ptr $ fromIntegral index
tensorflow                 >    |           ^^^
```

Minnozz:
Oops, thought I checked that. Fixed now.

tensorflow/src/TensorFlow/Internal/FFI.hs
```diff
@@ -122,43 +153,88 @@ shutDownRunner r = do
...
+graphImportGraphDef graph pb optionSetter =
+    useProtoAsBuffer pb $ \buffer ->
+        bracket Raw.newImportGraphDefOptions Raw.deleteImportGraphDefOptions $ \importGraphDefOptions -> do
+            optionSetter importGraphDefOptions
+            checkStatus $ Raw.graphImportGraphDef graph buffer importGraphDefOptions
+
+forGraphOperations_ :: Raw.Graph
+                    -> (Raw.Operation -> IO b)
+                    -> IO ()
+forGraphOperations_ graph f = with 0 go
```
fkm3, line 169: i think if the graph gets modified during the loop, we'd get unexpected results. can you add a note about that

ideally we'd keep the APIs with sharp edges contained in the `FFI` and `Raw` modules. one idea is to continue exporting an `extendGraph` function from `FFI`, just implement it using the logic you put into `Session`. not a blocker though

tensorflow/src/TensorFlow/Session.hs
```diff
@@ -117,17 +114,30 @@ sessionConfig = lens _sessionConfig (\g x -> g { _sessionConfig = x })
 sessionTracer :: Lens' Options Tracer
 sessionTracer = lens _sessionTracer (\g x -> g { _sessionTracer = x })
 
+-- | Run 'Session' actions in a new TensorFlow session.
+runSession :: (MonadMask m, MonadIO m) => SessionT m a -> m a
+runSession = runSessionWithOptions def
```
fkm3, line 119: can you move this back?

tensorflow/src/TensorFlow/Session.hs
```diff
@@ -28,7 +28,6 @@ module TensorFlow.Session (
     runSession,
     runSessionWithOptions,
     MonadBuild(..),
-    extend,
```
fkm3, line 31: do we need to stop exporting `extend`? this is part of the public API

tensorflow/src/TensorFlow/Session.hs
```diff
@@ -199,8 +217,8 @@ runWithFeeds_ feeds t = do
     ns <- build $ getNodes t
     runFetchWithFeeds feeds ns (pure ())
 
-fixFeeds :: [Feed] -> [(ByteString, FFI.TensorData)]
-fixFeeds = map $ \(Feed o d) -> (encodeUtf8 $ encodeOutput o, d)
+fixFeeds :: [Feed] -> [(String, FFI.TensorData)]
```
fkm3, line 220: is there a compelling reason to switch from `ByteString` to `String` throughout this PR? at least in the FFI code it is suspicious to me because were going to have to immediately pack the strings back into a contiguous array format for the C APIs to understand it

fkm3 (changes requested):
Thanks for figuring this out. I think a handful of people including myself have considered doing this and bounced off it.

tensorflow/src/TensorFlow/Session.hs
Minnozz: I wanted to use the `%` modifier in a [c2hs function hook](https://github.com/haskell/c2hs/wiki/Implementation-of-Haskell-Binding-Modules#function-hooks) because binding to functions with (non-pointer) struct parameters is not otherwise possible without a wrapper function. Switching to `String` seemed idiomatic because it appears in the list with default mashallers and `ByteString` does not, but I now see that it may be better to revert back to `ByteString` and specify the marshaller in the function hook.

tensorflow/src/TensorFlow/Session.hs
Minnozz: No, I thought it was more correct since using the builder and buffering the new nodes until we add them all at once is an implementation detail now, but I've added it back.

tensorflow/src/TensorFlow/Session.hs
```diff
@@ -139,16 +153,23 @@ instance Monad m => MonadBuild (SessionT m) where
...
 extend = do
     session <- Session (asks rawSession)
+    graph <- Session (asks rawGraph)
     trace <- Session (asks tracer)
     nodesToExtend <- build flushNodeBuffer
     unless (null nodesToExtend) $ liftIO $ do
         let graphDef = (defMessage :: GraphDef) & node .~ nodesToExtend
         trace ("Session.extend " <> Builder.string8 (showMessage graphDef))
-        FFI.extendGraph session graphDef
+        FFI.graphImportGraphDef graph graphDef $ \opts ->
```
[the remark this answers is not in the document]
Minnozz: `graphImportGraphDef` will error if a node has an input outside of the GraphDef, unless that node is mapped via an input mapping.

tensorflow/src/TensorFlow/Internal/FFI.hs
Minnozz: That sounds like a good idea; I dislike how much from `Internal` has to be exported to make writing the code in `Session` possible.

Minnozz: `TF_GraphNextOperation` does not mention anything about modification of the graph; should I add a comment anyway?

Minnozz (commented):
@fkm3 I think I've addressed your feedback, what do you think of the current PR?

tensorflow/src/TensorFlow/Internal/FFI.hs
fkm3: on second thought, i think we only really append to the graph, so there probably isn't much risk. i think this is fine

tensorflow/src/TensorFlow/Session.hs
fkm3: i see. i've forgotten most of how c2hs works. i remember avoiding `{# fun .. #}` for some reason, maybe just out of paranoia because it was hard to tell exactly what it was doing

fkm3 (approved):
looks good!

PR #289, opened 2023-02-16
Support Cabal 3.

mikesperber:
hsSourceDirs is now a list of SymbolicPath instead of FilePath.

This is needed to unbreak the Nix package, where Cabal 3 gets used.

### DIFF
--- a/tensorflow-core-ops/Setup.hs
+++ b/tensorflow-core-ops/Setup.hs
@@ -28,6 +28,9 @@ import Distribution.Simple
     , simpleUserHooks
     , UserHooks(..)
     )
+#if MIN_VERSION_Cabal(3,1,0)
+import Distribution.Utils.Path (unsafeMakeSymbolicPath)
+#endif
 import Data.List (intercalate)
 import Data.ProtoLens (decodeMessage)
 import System.Directory (createDirectoryIfMissing)
@@ -81,7 +84,11 @@ fudgePackageDesc lbi p = p
     }
   where
     fudgeBuildInfo bi =
+#if MIN_VERSION_Cabal(3,1,0)
+        bi { hsSourceDirs = unsafeMakeSymbolicPath (autogenModulesDir lbi) : hsSourceDirs bi }
+#else    
         bi { hsSourceDirs = autogenModulesDir lbi : hsSourceDirs bi }
+#endif
 
 blackList =
     [ -- Requires the "func" type:


tensorflow-core-ops/Setup.hs
```diff
@@ -81,7 +84,11 @@ fudgePackageDesc lbi p = p
     }
   where
     fudgeBuildInfo bi =
+#if MIN_VERSION_Cabal(3,0,0)
+        bi { hsSourceDirs = unsafeMakeSymbolicPath (autogenModulesDir lbi) : hsSourceDirs bi }
```
fkm3, line 88: it is odd that there isn't an obvious way to create a `SymbolicPath` except by using this "unsafe" function

tensorflow-core-ops/Setup.hs
```diff
@@ -28,6 +28,9 @@ import Distribution.Simple
     , simpleUserHooks
     , UserHooks(..)
     )
+#if MIN_VERSION_Cabal(3,0,0)
+import Distribution.Utils.Path (unsafeMakeSymbolicPath, SymbolicPath)
```
fkm3, line 32: probably need to remove the `SymbolicPath` import to avoid an "unused import" warning

tensorflow-core-ops/Setup.hs
mikesperber: I agree. Looked for a while but didn't see any other way to do it.

tensorflow-core-ops/Setup.hs
mikesperber: Good point, done.

fkm3:
The build server is getting this error:

```
tensorflow-core-ops        > /[tfhs/tensorflow-core-ops/Setup.hs:32](https://cs.corp.google.com/piper///depot/google3/tfhs/tensorflow-core-ops/Setup.hs?l=32):1: error:
tensorflow-core-ops        >     Could not find module ‘Distribution.Utils.Path’
tensorflow-core-ops        >     Perhaps you meant
tensorflow-core-ops        >       Distribution.Utils.IOData (from Cabal-3.0.1.0)
tensorflow-core-ops        >       Distribution.Utils.Base62
tensorflow-core-ops        >       Distribution.Utils.String
tensorflow-core-ops        >     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
tensorflow-core-ops        >    |
tensorflow-core-ops        > 32 | import Distribution.Utils.Path (unsafeMakeSymbolicPath, SymbolicPath)
tensorflow-core-ops        >    | ^
```

Seems like we've got `Cabal-3.0.1.0` and the old code works with it. Maybe it isn't Cabal 3 that is the problem, but a later version and you need to raise the `MIN_VERSION_Cabal`?

mikesperber:
> Seems like we've got `Cabal-3.0.1.0` and the old code works with it. Maybe it isn't Cabal 3 that is the problem, but a later version and you need to raise the `MIN_VERSION_Cabal`?

Ah, it looks like it was done in Cabal 3.1.0:

https://github.com/haskell/cabal/commit/18459e7b9ea404940b6d598c5a033724352051df
 
Sorry about that!

@fkm3 I updated the patch - could you try again?

fkm3 (approved):

fkm3:
seems good now, thanks

PR #292, opened 2023-08-10
Update to TensorFlow 2.12

jbransen:
Commit https://github.com/tensorflow/tensorflow/commit/78d8fdde71b9a0dc2e1919194f071ba8c222e1d5 renamed the status fields from the `TF_` prefix to `TSL_`, though there are define's in place to keep the old names, the Haskell FFI generator picks up the new names, leading to compile errors.

To fix compilation, and be forward compatible, I renamed the fields, which means this branch is not compatible with older tensorflow versions, but I don't know how to make it so without ugly hacks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ working, the following commands will compile and run the tests.
 ```
 git clone --recursive https://github.com/tensorflow/haskell.git tensorflow-haskell
 cd tensorflow-haskell
-docker build -t tensorflow/haskell:2.10.1 docker
+docker build -t tensorflow/haskell:2.12.0 docker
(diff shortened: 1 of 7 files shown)


blackgnezdo:
Do you feel like updating the pinned version of TF at the same time? Then you don't have to worry about backward compatibility. Some of the recent commits are such updates, so you have a template to start with.

jbransen:
Yes, that makes sense, I believe I did that now (though I did not actually run the Docker containers, but I did verify that the base images exist so I have good hopes that this is right).

fkm3 (approved):

fkm3:
Thanks!

